PR #9, opened 2022-07-28
feat(imagery): Non Visual QA workflow-template TDE-309

paulfouquet:
## Description
In order to use the `non_visual_qa.py` script from `Topo-Imagery`, we need a workflow template that allows us to submit a workflow from Argo.

## Change
Add the workflow template `non-visual-qa.yaml`

## Tests
The workflow template has been created in Argo `imagery-non-visual-qa-l9z66` under the namespace `argo`.

### DIFF
--- a/workflows/imagery/non-visual-qa.yaml
+++ b/workflows/imagery/non-visual-qa.yaml
@@ -3,6 +3,8 @@ apiVersion: argoproj.io/v1alpha1
 kind: WorkflowTemplate
 metadata:
   name: imagery-non-visual-qa-v0.2.0-20
+  annotations:
+    fluentbit.io/parser: json
 spec:
   serviceAccountName: workflow-runner-sa
   podGC:
--- a/workflows/imagery/non-visual-qa.yaml
+++ b/workflows/imagery/non-visual-qa.yaml
@@ -54,6 +54,8 @@ spec:
             "list",
             "--filter",
             "{{inputs.parameters.filter}}",
+            "--group",
+            "10",
             "--output",
             "/tmp/file_list.json",
             "--config",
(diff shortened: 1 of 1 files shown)


workflows/imagery/non-visual-qa.yaml
```diff
@@ -0,0 +1,85 @@
...
+                  value: "{{item}}"
+            depends: "aws-list"
+            withParam: "{{tasks.aws-list.outputs.parameters.files}}"
+    - name: aws-list
+      inputs:
+        parameters:
+          - name: uri
+          - name: filter
+      container:
+        resources:
```
blacha, line 45: this container doesnt need anywhere near this amount of resources.

workflows/imagery/non-visual-qa.yaml
```diff
@@ -0,0 +1,85 @@
+---
+apiVersion: argoproj.io/v1alpha1
+kind: WorkflowTemplate
+metadata:
+  generateName: imagery-non-visual-qa-
```
blacha, line 5: if we leave this as generate name it will  create a new name everytime we deploy a new workflow template

paulfouquet: The current way of generating a name is not ideal, but is that not a generated name that we want to manage versioning?

workflows/imagery/non-visual-qa.yaml
paulfouquet: What resources do you suggest?

workflows/imagery/non-visual-qa.yaml
paulfouquet: Resolved using container version number

blacha (approved):

PR #14, opened 2022-08-12
feat(workflow): Create config file after standardise step finished.

Wentao-Kuang:


workflows/imagery/standardising.yaml
```diff
@@ -91,6 +105,50 @@ spec:
             path: /tmp/
             archive:
               none: {}
+    - name: get-bucket
+      container:
+        image: stedolan/jq:latest
+        command: [sh, -c]
+        args:
```
blacha, line 112: could this just be one workflow with one output of the location? you can use just straight python or node?

eg 

```typescript
const fs = require('fs');
const loc = JSON.parse(process.env['ARGO_TEMPLATE']).archiveLocation.s3;
writeFileSync("/tmp/location", `${loc.bucket}/${loc.key}`
```

Wentao-Kuang: That's awesome, I didn't know that we can directly run js in workflow. Going to following this example, https://github.com/argoproj/argo-workflows/blob/master/examples/scripts-javascript.yaml

Wentao-Kuang: Updated. Looks like working fine. 
https://argo.linzaccess.com/workflows/argo/imagery-standardise-v0.2.0-20-config-nc5nb?tab=workflow&nodeId=imagery-standardise-v0.2.0-20-config-nc5nb-1924812991&nodePanelView=inputs-outputs

workflows/imagery/standardising.yaml
```diff
@@ -91,6 +100,37 @@ spec:
...
               none: {}
+    - name: get-location
+      script:
+        image: node:alpine
+        command: [node]
+        source: |
+          const fs = require('fs');
+          const loc = JSON.parse(process.env['ARGO_TEMPLATE']).archiveLocation.s3;
+          const keys = loc.key.split('/');
+          fs.writeFileSync('/tmp/location', `s3://${loc.bucket}/${keys[0]}/${keys[1]}/`);
```
blacha, line 111: this is not super portable, could you just use `dirname()` or similar to remove the last entry.

If  we change how deep this key is it will break.

Wentao-Kuang: Sure, this logic is just remove the pod name at the end of path. 

If we change the key format for `artifactRepository`, this might broken. 
https://github.com/linz/topo-aws-infrastructure/blob/9924a72ff88bf2d1a7fd208a2bb7348e9402b300/src/stacks/eks/eks.workflow.ts#L65

blacha: you could also just replace off the podname which should also be in the `ARGO_TEMPLATE` or `ARGO_NODE_ID` or something?

Wentao-Kuang: Fixed by removing the `pod.name` in the path. 
https://argo.linzaccess.com/workflows/argo/imagery-standardise-v0.2.0-20-config-qkghn?tab=workflow&nodeId=imagery-standardise-v0.2.0-20-config-qkghn-663116485&nodePanelView=inputs-outputs

blacha (approved):

PR #16, opened 2022-08-16
feat: workflow that sleeps for a hour for testing

blacha:
developers can submit the workflow then use kubectl to connect to the machine.

```
argo submit workflows/test/sleep.yml
k exec -it sleep-:podName /bin/bash
```


paulfouquet (approved):
:zzz:

PR #20, opened 2022-08-24
feat(workflow): New workflow to create mapsheet from provided layer id.

Wentao-Kuang:


workflows/imagery/cog-mapsheet.yaml
```diff
@@ -0,0 +1,107 @@
...
+    - name: fetch-layer
+      inputs:
+        parameters:
+          - name: layer
+      container:
+        image: ghcr.io/linz/argo-tasks:v1.0.0-8-g53a2c88
+        command: [node, index.cjs]
+        env:
+          - name: AWS_ROLE_CONFIG_PATH
+            value: s3://linz-bucket-config/config-v2.json
```
blacha, line 49: this doesnt need to be `-v2.json` any more.

workflows/imagery/cog-mapsheet.yaml
```diff
@@ -0,0 +1,107 @@
...
+          - name: geopackage
+            path: /tmp/geopackage.gpkg
+
+    - name: transform
+      inputs:
+        artifacts:
+          - name: geopackage
+            path: /tmp/geopackage.gpkg
+      container:
+        image: osgeo/gdal:alpine-small-latest
```
blacha, line 69: I think we should avoid `-latest` tags, can you use `-3.5.0`

Wentao-Kuang: Cool, I will try with that to see if this works. Just some old version might not have `FlatGeobuf` driver.

Wentao-Kuang: All fixed, and tested with alpine-small-3.5.0 works fine. Thanks.

blacha (approved):

PR #22, opened 2022-08-24
feat(workflow): Move the create-mapsheet workflow into basemaps workflow.

Wentao-Kuang:


workflows/basemaps/cog-mapsheet.yaml
```diff
@@ -0,0 +1,105 @@
...
+kind: WorkflowTemplate
+metadata:
+  name: create-cog-mapsheet
+  namespace: argo
+spec:
+  entrypoint: main
+  arguments:
+    parameters:
+      - name: layer
+        value: "104687"
```
paulfouquet, line 12: Is that a default value that you want to leave?

Wentao-Kuang: Yes, it is.

paulfouquet (approved):

workflows/basemaps/cog-mapsheet.yaml
blacha: I wonder if there is a better way to handle these `layerIds` ? to be fair this specific `layerId`  wont ever change?

Wentao-Kuang: I think we can change this to anything when we submit the job. This is just the default layer id.

PR #29, opened 2022-08-29
ci: deploy workflow templates on merge to master

blacha:


.github/workflows/main.yml
```diff
@@ -0,0 +1,76 @@
...
+        run: |
+          aws eks update-kubeconfig --name Workflow --region ap-southeast-2 --role-arn ${{ secrets.AWS_EKS_ROLE }}
+
+      - name: Check EKS connection
+        run: |
+          kubectl get nodes
+
+      - name: Install Argo
+        run: |
+          curl -sLO https://github.com/argoproj/argo-workflows/releases/download/v3.4.0-rc2/argo-linux-amd64.gz
```
blacha, line 59: should we be running a RC variant!? I couldnt find a `setup-argo` workflow.

.github/workflows/main.yml
```diff
@@ -0,0 +1,76 @@
...
+          gunzip argo-linux-amd64.gz
+          chmod +x argo-linux-amd64
+          ./argo-linux-amd64 version
+
+      - name: Lint workflows
+        run: |
+          ./argo-linux-amd64 lint workflows/
```
blacha, line 66: `argo lint` requires a connection to the k8s cluster :sad-panda:

.github/workflows/main.yml
```diff
@@ -0,0 +1,76 @@
...
+
+      - name: Deploy workflows
+        run: |
+          # Find all workflows that have kind "WorkflowTemplate"
+          WORKFLOWS=$(grep '^kind: WorkflowTemplate$' -R workflows/ -H | cut -d ':' -f1)
+
+          # For each workflow attempt to deploy it using kubectl
+          for wf in $WORKFLOWS; do
+              kubectl apply -f $wf --namespace argo
```
blacha, line 75: I wonder if there is a better way to do this... ? open to suggestions.

.github/workflows/main.yml
paulfouquet: I don't think it is critical for this

workflows/imagery/non-visual-qa.yaml
```diff
@@ -3,6 +3,7 @@ apiVersion: argoproj.io/v1alpha1
 kind: WorkflowTemplate
 metadata:
   name: imagery-non-visual-qa-v0.2.0-34
+  namespace: argo
```
paulfouquet, line 6: Should we set this as default in https://github.com/linz/topo-aws-infrastructure/blob/7b1aec4f9030ded543b074a0e980dfbf7ed788d9/src/stacks/eks/eks.workflow.ts#L96 ?

blacha: I don't think `kubectl` respect those defaults.

paulfouquet (approved):

MDavidson17 (approved):

PR #34, opened 2022-10-06
feat: stac-validate workflow TDE-499

amfage:
The output does not display very nicely in Argo, so there may be a better way to do that.

image


MDavidson17 (approved):
:+1: Looks great to me.
If changing the look of the output is a big job maybe it could be a separate task?

PR #44, opened 2022-11-01
feat(basemaps): imagery import workflow for make cog.

Wentao-Kuang:


paulfouquet (approved):
The workflow looks good to me

PR #54, opened 2022-11-11
feat(basemaps): Create basemaps overview cogs workflow.

Wentao-Kuang:
…rge cli changes.


workflows/basemaps/create-overview.yaml
```diff
@@ -0,0 +1,53 @@
...
+        command: [node, index.cjs]
+        env:
+          - name: AWS_ROLE_CONFIG_PATH
+            value: s3://linz-bucket-config/config.basemaps.json
+        args:
+          [
+            "-V",
+            "create-overview",
+            "--source",
+            "{{inputs.parameters.source}}",
```
paulfouquet, line 48: You can pass directly `workflow.parameters.*` if you'd like. That would avoid you to link the arguments in your task above (in `template.dag.tasks.arguments.parameters`)

paulfouquet: One downside is that you pod summary does not reference these input values in the UI. But you can still see them in the main Workflow summary in Argo UI.

paulfouquet (dismissed):

workflows/basemaps/create-overview.yaml
Wentao-Kuang: I think we can keep inputs for now? Because eventually this will be running as a part of import process like create-config. So the inputs might come from another task.

paulfouquet (approved):

PR #62, opened 2022-11-21
docs: stac-validate README TDE-572

amfage:


workflows/stac/README.md
```diff
@@ -0,0 +1,42 @@
...
+
+If any schema validation failures are found, the workflow will fail.
+
+Logs can be viewed in Argo by selecting the `stac-validate` workflow component > CONTAINERS > LOGS.
+
+![WorkflowLogs](../../docs/workflow_logs.png)
+
+Example failure messages:
+
+```bash
```
dwsilk, line 26: use `json` rather than `bash`? Would be nice for this to be properly formatted but we don't get a formatted output at the moment so I guess that would be inaccurate.

dwsilk (dismissed):

workflows/stac/README.md
amfage: I think `json` is better as it's fundamentally what you see in Argo just with some syntax highlighting.

PR #68, opened 2022-11-30
feat: tileset-validate standalone workflow TDE-586

amfage:


workflows/imagery/tileset-validate.yaml
```diff
@@ -0,0 +1,44 @@
...
+          - name: tileset-validate
+            template: tileset-validate
+            arguments:
+              parameters:
+                - name: processed-imagery-path
+                  value: "{{workflow.parameters.processed-imagery-path}}"
+              artifacts:
+                - name: tileset-validate-result
+                  raw:
+                    data: "{ { tasks.tileset-validate.outputs.result } }"
```
paulfouquet, line 28: Is it the auto-formatting that did format this value like that? (spaces)

amfage: It was also like that in `stac-validate`, I've fixed them.

paulfouquet (approved):

PR #74, opened 2022-12-20
feat: make flatten optional in publish copy

MDavidson17:
In order to move the data from `linz-imagery-upload` to `linz-topographic-upload`

Tested successfully by copying all the existing data from linz-imagery-upload to linz-topographic-upload

This change will affect TDP so they should be warned of the changes and what it means for their workflows (run with --flatten option)


MDavidson17:
Decided this change is not required in production as it was made for one-off use in a specific situation, so will close pull request but leave branch.

PR #80, opened 2023-01-18
feat: limit standardising workflow concurrency using semaphore

ajacombs:
This sets a limit (initially set at **2**) on how many instances of the standardising workflow can be run concurrently. This will allow us to bulk submit workflows for the historical imagery standardising and not cause issues by having too many workflows run at the same time.

Changes made in this PR:
- Adds a new config directory, with a [semaphores.yml](config/semaphores.yml) file containing a k8s ConfigMap containing the synchronization semaphore value.
- References the ConfigMap name and key from the standardising workflow template.
- Increments the workflow template version number.
- Updates the GitHub Actions workflow to deploy any yaml files in the `config` directory to k8s using `kubectl apply`.
- Adds docs outlining how to use the synchronization feature.


workflows/imagery/standardising.yaml
```diff
@@ -114,6 +114,11 @@ spec:
                   value: "{{tasks.get-location.outputs.parameters.location}}"
             template: create-config
             depends: "get-location && create-overview"
+      synchronization:
```
blacha, line 117: Where is this synchronization applied at the DAG level? so the entire DAG is limited to one?

paulfouquet (dismissed):
Looks good to me. As you mentioned, let see how it behaves in the UI with the Argo upgrade

ajacombs:
I've updated this to use workflow-level synchronization, whereas previously I had used template-level synchronization on the `main` template within the workflow.

Both approaches have the same effect: restricting how many instances of the `main` template can run is equivalent to restricting how many instances of the workflow itself can run.

However workflow-level synchronization has the advantage that workflows are shown as pending in the Argo GUI, which will allow users to see the progress of a bulk lot of workflows.

workflows/imagery/standardising.yaml
ajacombs: Updated with new commit to change to workflow-level synchronization.

blacha (approved):

ajacombs:
Workflows pending because of workflow-level synchronization were previously not visible in the Argo GUI due to [a bug in Argo](https://github.com/argoproj/argo-workflows/issues/9936) which was present in our deployed version of 3.4.3.

Upgrading to Argo 3.4.4 has resolved this, and a test using the [test.yml](workflows/test/sleep.yml) workflow successfully saw the pending workflows display correctly in the Argo GUI.

PR #81, opened 2023-01-19
feat(basemaps): Add concurrency and category to the imagery import process

Wentao-Kuang:


Wentao-Kuang:
Test pr for category

workflows/basemaps/imagery-import.yaml
```diff
@@ -149,10 +160,10 @@ spec:
           - name: job
           - name: names
       container:
-        image: ghcr.io/linz/basemaps/cli:v6.36.0-33-gc50b85cc
+        image: ghcr.io/linz/basemaps/cli:v6.38.0-23-gd4715ac0
         resources:
           requests:
-            memory: 7.8Gi
+            memory: 10Gi
```
blacha, line 166: do we need more memory? using round numbers here is bad as it means things cannot easily fit on machines due to overheads.

generally everything is increments of 8Gi so if you wanted this bigger 15.6Gi would be the next size up

Wentao-Kuang: Ok, i just worried that it will die with 8GB, I will revert this back and increase to 15.6 Gi if this failed in future.

Wentao-Kuang: Updated.

paulfouquet (dismissed):

amfage (approved):

PR #83, opened 2023-01-20
fix: standardise parameter names

blacha:


MDavidson17 (approved):

amfage (dismissed):

amfage:
I just realised that the README also needs to be updated for the parameter name change.

paulfouquet (approved):

PR #84, opened 2023-01-20
feat: manage container version from workflow parameters TDE-527

paulfouquet:
- Manage LINZ container versions using a workflow parameter
- Set `ImagePullPolicy` to ensure image is pulled to the latest match to the tag set in parameter
- Start versioning `Workflow` and `WorkflowTemplate` independently from containers
- Update docs


workflows/basemaps/create-overview-all.yaml
```diff
@@ -2,13 +2,17 @@
 apiVersion: argoproj.io/v1alpha1
 kind: Workflow
 metadata:
-  name: create-all-overviews
+  name: create-all-overviews-v1-
```
amfage, line 5: Some of the workflow names have a trailing `-`, but not all.

paulfouquet: That is a good point. There is a difference between `Workflow` and `WorkflowTemplate`: the generated name for `WorkflowTemplate` automatically adds a trailing `-` whereas it does not for `Workflow`.

amfage (dismissed):

amfage (approved):

PR #100, opened 2023-02-13
feat: standardise-publish-import workflow TDE-627

amfage:


workflows/imagery/standardising-publish-import.yaml
```diff
@@ -0,0 +1,72 @@
...
+        value: "v6"
+      - name: version-topo-imagery
+        value: "v0"
+  templateDefaults:
+    container:
+      imagePullPolicy: Always
+  templates:
+    - name: main
+      steps:
+        - - name: standardise
```
paulfouquet, line 41: Is it normal to have two `-` here?

amfage: Single dash is for running in parallel and double dash means they will run sequentially. There is an example here: https://argoproj.github.io/argo-workflows/walk-through/steps/

paulfouquet (dismissed):

paulfouquet (dismissed):

paulfouquet (approved):

PR #101, opened 2023-02-15
feat: optional parameter licensor-list TDE-643

paulfouquet:
## Description
Add a new parameter `licensor-list` that allows the user to pass multiple STAC `licensor` names to create the `collection.json`.
image
This parameter is optional and does not have a control on the names passed. The list has to be separated by semicolon `;` and takes effect only if there are more than 1 element, otherwise it's `licensor` (the dropdown list) that is used.
Related to https://github.com/linz/topo-imagery/pull/355

### DIFF
--- a/workflows/imagery/standardising.yaml
+++ b/workflows/imagery/standardising.yaml
@@ -163,7 +163,7 @@ spec:
             "Whangārei District Council",
           ]
       - name: licensor-list
-        description: "(Optional) Separate the licensor names by a semi-colon (;). It has no affect unless a semicolon-delimited list is entered."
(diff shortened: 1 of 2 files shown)


workflows/imagery/standardising.yaml
```diff
@@ -162,6 +162,9 @@ spec:
             "Whanganui District Council",
             "Whangārei District Council",
           ]
+      - name: licensor-list
+        description: "(Optional) Separate the licensor names by a semi-colon (;). It has no affect unless a semicolon-delimited list is entered."
```
amfage, line 166: NIT: "effect"

MDavidson17 (approved):

PR #102, opened 2023-02-16
Feat: tool generate params TDE-632

MDavidson17:
nb: this branch may not be merged however the code will remain here fore the time being

**to run:**

1. download the imagery [parameters csv](https://linzsrm.sharepoint.com/:x:/r/sites/Topography/_layouts/15/Doc.aspx?sourcedoc=%7B508567E2-EF88-458B-9115-0FC719CAA540%7D&file=imagery-standardising-parameters-bulk-process.xlsx&action=default&mobileredirect=true) and convert to a csv
2. save the csv in the `tools` directory, if you change the name/location you will need to change the `PARAMETERS_CSV` field
3. run:
```bash
cd tools
python3 ./generate-argo-cli-commands.py | pjl
```
**Output:**
- numerous parameter yaml files to be processed
- a sh script containing all the argo commands which need to be run
- logs detailing:
    - any datasets which also need to be imported into basemaps
    - any datasets which have been skipped and the reasons why

### DIFF
--- a/tools/README.md
+++ b/tools/README.md
@@ -1,0 +1,14 @@
+# Tools
+
+This folder contains single use scripts which have been used to assist in running argo workflows. 
+The scripts are stored in this folder if it is thought they may become useful again in the future. 
+
+## generate-argo-commands-imagery.py
+**Date:** 14/02/2023
+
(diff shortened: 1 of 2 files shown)


paulfouquet (commented):
LGTM

MDavidson17:
Example output:
[tools.zip](https://github.com/linz/topo-workflows/files/10760643/tools.zip)

MDavidson17:
updates based on reviews:

- `Some of the source path end with / some other without. I think aws-list handles that` - trailing / stripped
- `The description shows up with 2 spaces on the line return, I don't know if there is an impact?` - yarn.dumps has line length parameter - updated to 1000
- `bay-of-plenty-0_25m-rural-aerial-photos-(2011-2012).yaml for example referes BOPLASS Limited instead of BOPLASS in the licensor-list param` - fixed is original spreadsheet 
- licensor added as empty if licensor-list
- changed filenames to match target name
- should we wrap the values in the yaml file in "" (WIP)

PR #121, opened 2023-04-13
feat: bulk stac-validate collections TDE-504

amfage:


workflows/stac/stac-validate.yaml
```diff
@@ -5,45 +5,122 @@ metadata:
...
           - name: AWS_ROLE_CONFIG_PATH
             value: s3://linz-bucket-config/config.json
         args:
           [
             "stac-validate",
+            "--concurrency",
+            "120",
             "--recursive",
-            "{{=sprig.trim(inputs.parameters['stac-file-path'])}}",
+            "--checksum",
```
blacha, line 124: Are we able to do something like `--checksum={{{{workflow.parameters.checksum}} == false}}`

blacha: actually, it only needs true or false here, so `--checksum={{workflow.parameters.checksum}}` could work?

MDavidson17 (dismissed):

PR #128, opened 2023-05-03
feat(basemaps): new workflow for the vector map data etl. BM-804

Wentao-Kuang:


workflows/basemaps/vector-etl.yaml
```diff
@@ -0,0 +1,86 @@
...
+          - name: layer
+            valueFrom:
+              path: "/tmp/layer.json"
+
+    - name: create-pr
+      inputs:
+        parameters:
+          - name: layer
+      container:
+        image: ghcr.io/linz/basemaps/cli:v6.40.0-25-g637df773
```
blacha, line 70: this is a very specific commit maybe `:latest` or we will need to change this often?

Wentao-Kuang: Yeah, I think this won't be changed often. I mark as latest just in case Karl want to add some new labels or change zoom levels for some layers.

Wentao-Kuang: Updated.

blacha (approved):

PR #133, opened 2023-05-23
feat: add basemaps-imagery-import-cogify workflow BM-858

blacha:
#### Description

Adds a new Argo workflow template to import imagery to Basemaps using the new cogify commands.

#### Intention

This will allow imagery to be imported to Basemaps using the improved cogify commands, which addresses slight resampling errors (the 'faint white lines' bug) and improves the efficiency of the size of COGs generated.

#### Checklist

- [ ] Tests updated. N/A: no automated tests for Argo workflows. Has however been [manually tested successfully with historical imagery data](https://argo.linzaccess.com/workflows/argo/test-cogify-dkplw).
- [x] Docs updated
- [x] Issue linked in title

### DIFF
--- a/workflows/basemaps/imagery-import-cogify.yml
+++ b/workflows/basemaps/imagery-import-cogify.yml
@@ -13,11 +13,11 @@ spec:
     parameters:
       - name: version_basemaps_cli
         description: Version of the basemaps CLI docker container to use
-        value: latest
+        value: v6
 
       - name: version_argo_tasks
         description: Version of the basemaps CLI docker container to use
-        value: latest
+        value: v2
 
       - name: source
         description: Source imagery location "s3://linz-imagery"
--- a/workflows/basemaps/imagery-import-cogify.yml
+++ b/workflows/basemaps/imagery-import-cogify.yml
@@ -1,7 +1,7 @@
 apiVersion: argoproj.io/v1alpha1
-kind: Workflow
+kind: WorkflowTemplate
 metadata:
-  generateName: test-cogify-
+  generateName: basemaps-imagery-import-cogify-
   namespace: argo
 spec:
   parallelism: 100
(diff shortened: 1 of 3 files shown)


workflows/test/test-cogify.yml
```diff
@@ -0,0 +1,264 @@
...
+      
+      - name: version_argo_tasks
+        description: Version of the basemaps CLI docker container to use
+        value: latest
+
+      - name: source
+        description: Source imagery location "s3://linz-imagery"
+      
+      - name: target
+        description: Target location for output COGs
```
Wentao-Kuang, line 23: This could be enum for the available target bucket?
`value: "linz-basemaps"
        enum:
          - "linz-basemaps"
          - "linz-basemaps-staging"`

ajacombs: Added this to workflow

workflows/basemaps/imagery-import-cogify.yml
```diff
@@ -0,0 +1,279 @@
...
+  parallelism: 100
+  entrypoint: main
+  templateDefaults:
+    container:
+      imagePullPolicy: Always
+  arguments:
+    parameters:
+      - name: version_basemaps_cli
+        description: Version of the basemaps CLI docker container to use
+        value: latest
```
paulfouquet, line 16: Is that intended to set the `latest` version by default?

blacha: As this is still somewhat a testing workflow, we are mostly running directly off master for it.

ajacombs: Nope, good spotting! :smile: We had them as `latest` for testing but they should be set to the current major version now it's moving out of testing. Have pushed a new commit which does that.

paulfouquet (approved):

workflows/basemaps/imagery-import-cogify.yml
blacha: I think you are going to want to leave it on latest, our release cycle means we don't really release that major version very often, I don't think the latest major version is actually stable for this either.

PR #138, opened 2023-07-04
feat: add parameter to optionally skip tile-index-validate step

ajacombs:
Adds a parameter to the standardising workflow to allow optionally skipping the tile-index-validate step. The default is to not skip.

Currently some datasets (such as non-geotiff .tif files with .tfw or .prj sidecars) fail this step, but are able to be successfully standardised as later steps using GDAL parse these sidecars happily.

Adding a parameter to allow skipping the tile-index-validate step would remove the need to submit jobs using the separate standardising-no-epsg workflow template, and to maintain that template as changes are made to the standardising template.


paulfouquet (approved):

amfage:
Not required any more, will fix the EPSG code issue in argo-tasks in TDE-774

PR #140, opened 2023-07-05
feat: preset and producers/licensors for elevation TDE-781

amfage:
This is in draft until the `topo-imagery` changes are made to handle retiling and the new output from `argo-tasks` `tileindex-validate`.

The Basemaps steps will fail until Basemaps can handle the elevation output.


amfage:
Incorporated into https://github.com/linz/topo-workflows/pull/142

PR #162, opened 2023-08-28
feat: add slack notification onExit TDE-842

paulfouquet:
## Description
Trigger a Slack notification at the end of the create thumbnails workflow to post on `#alert-argo-workflows`.

image

### DIFF
--- a/workflows/util/README.md
+++ b/workflows/util/README.md
@@ -33,6 +33,8 @@ graph TD;
     create-thumbnails-->publish-thumbnails;
 ```
 
+Note: `#alert-argo-workflows` Slack channel is notified once the workflow is completed.
+
 ### [aws-list](https://github.com/linz/argo-tasks/blob/master/src/commands/list/list.ts)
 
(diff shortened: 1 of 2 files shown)


MDavidson17 (approved):

blacha:
Before we merge this, I think we need to decide *if* we want to move our slack messaging from elastic into argo.

I think we need a bit of a pros/cons on if we think Argo is a better place for alerts? then also if we do decide to go with argo, some sort of migration path/plan for all the existing alerts?

paulfouquet:
> Before we merge this, I think we need to decide _if_ we want to move our slack messaging from elastic into argo.
> 
> I think we need a bit of a pros/cons on if we think Argo is a better place for alerts? then also if we do decide to go with argo, some sort of migration path/plan for all the existing alerts?

Convert this PR to a draft until we talk about this.

paulfouquet:
Closing this as our alerting system is managed by Elasticsearch watchers for now.

PR #170, opened 2023-09-07
feat: approval step in publish-copy TDE-853

MDavidson17:
**Spike Findings:**
**Filtering:**
- The approval step can be filtered to only run when the target starts with `s3://linz-imagery` / `s3://linz-elevation` /`s3://nz-imagery/`
**Parameters:**
- The approval step could allow the user to correct the target / reconfirm the target

**UI**
- It is not clear that approval is required when looking at the running workflows in list view
- It is clear a workflow needs approval when viewing the specific workflow

**Timeout**
- Can add a duration step/task to automatically resume the workflow after a certain period of time (somewhat defeats the purpose in our case?)
- If you want to abort a workflow after a certain duration of time if not approved you have to use the following workaround at the moment: take user confirmation (`YES`/`NO`) defaulted to `NO` add a timeout of 1hr if `parameter = NO` run `abort workflow` task (https://github.com/argoproj/argo-workflows/discussions/8365#discussioncomment-6419368).

**CLI Resume Commands:**
```bash
# Resume a workflow that has been suspended:

  argo resume *wf*

# Resume multiple workflows:

  argo resume *wf* *wf-1* *wf-2*     

# Resume the latest workflow:

  argo resume @latest

# Resume Workflow with Change to Target:
# TODO: unsure if possible to specify paramters from CLI https://argoproj.github.io/argo-workflows/intermediate-inputs/
```

**Docs:**
[Suspending Workflows](https://argoproj.github.io/argo-workflows/walk-through/suspending/)
[Resuming Workflows](https://argoproj.github.io/argo-workflows/cli/argo_resume/)

**UI Screenshots:**

Workflow structure example (skipping approval):
image

Approval Message:
(If the target is re-typed in the dialog box the new target will be used)
workflow

### DIFF
--- a/workflows/imagery/publish-copy.yaml
+++ b/workflows/imagery/publish-copy.yaml
@@ -49,6 +49,10 @@ spec:
           - name: group-size
       dag:
         tasks:
+          - name: approval
+            template: approval
+            when: "{{=sprig.regexMatch('s3://linz-(elevation|imagery)/', workflow.parameters.target)}}"
+
           - name: create-manifest-github
             template: create-manifest
             arguments:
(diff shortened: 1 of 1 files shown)


MDavidson17:
not implementing this solution as it is not quite fit for our purpose, will instead using github PR and Actions to submit a workflow.

PR #172, opened 2023-09-12
feat: format (pretty-print) STAC files TDE-759

paulfouquet:
### Description
Add a `WorkflowTemplate` to format files that calls `linz/argo-tasks` `pretty-print` command. This template is used in `imagery-standardising` workflow.

### Intention
Some of the STAC files (`Item`) are not pretty printed. Having a `WorkflowTemplate` allows to format files inside a workflow for newly created files. It can also be ran as a one-off job for existing files.

### Checklist
If not applicable, provide explanation of why.

- [x] Docs updated
- [x] Issue linked in Title

### DIFF
--- a/templates/argo-tasks/format.yml
+++ b/templates/argo-tasks/format.yml
@@ -1,0 +1,32 @@
+apiVersion: argoproj.io/v1alpha1
+kind: WorkflowTemplate
+metadata:
+  name: tpl-at-format
+spec:
+  templateDefaults:
+    container:
+      imagePullPolicy: Always
+  entrypoint: main
+  templates:
+    - name: main
+      inputs:
+        parameters:
+          - name: source
+            description: Path to format files from
+          - name: target
+            description: (Optional) Target directory (testing)
+            default: ""
+          - name: version
+            description: argo-task Container version to use
+            default: "v2"
+
+      container:
+        image: "019359803926.dkr.ecr.ap-southeast-2.amazonaws.com/eks:argo-tasks-{{= inputs.parameters.version }}"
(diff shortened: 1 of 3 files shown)


templates/argo-tasks/format.yml
```diff
@@ -0,0 +1,36 @@
+apiVersion: argoproj.io/v1alpha1
+kind: WorkflowTemplate
+metadata:
+  name: tpl-at-format
```
l0b0, line 4: What's "at" in this? Would "stac" be clearer?

templates/argo-tasks/format.yml
```diff
@@ -0,0 +1,36 @@
...
+      imagePullPolicy: Always
+  entrypoint: main
+  templates:
+    - name: main
+      inputs:
+        parameters:
+          - name: source
+            description: Path to format files from
+          - name: target
+            description: (Optional) Target directory (testing)
```
l0b0, line 17: If this is empty, does it format files in-place? As a newbie, that might be useful clarification.

templates/argo-tasks/format.yml
```diff
@@ -0,0 +1,36 @@
...
+            default: ""
+          - name: fix-content-type
+            description: (Optional) Set the content-type to S3 object
+            default: ""
```
l0b0, line 21: What are valid values here? Are these [MIME types](https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types), as used by the HTTP `Content-Type` header? What happens if this is empty?

PR #183, opened 2023-09-28
fix: add source to workflow

MDavidson17:
The merging of https://github.com/linz/topo-workflows/pull/161 linted against the old version of standardising so missed this required fix which broke Wentao's PR here: https://github.com/linz/topo-workflows/actions/runs/6333551973/job/17201837027?pr=179

### DIFF
--- a/workflows/imagery/standardising-publish-import.yaml
+++ b/workflows/imagery/standardising-publish-import.yaml
@@ -16,6 +16,8 @@ spec:
         key: bulk
   arguments:
     parameters:
+      - name: source
+        value: ""
       - name: cutline # optional standardising cutline
         value: ""
       - name: collection-id # optional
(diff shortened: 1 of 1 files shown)


paulfouquet (approved):

workflows/imagery/standardising-publish-import.yaml
```diff
@@ -16,6 +16,8 @@ spec:
         key: bulk
   arguments:
     parameters:
+      - name: source
+        value: ""
```
Wentao-Kuang, line 20: Do we need to put a default value here as this is not optional?

MDavidson17: oh sorry, I tried to remove it from the merge queue when I saw your comment but I was too slow, I will look into it and update in a new PR :joy: but it should work for you now

PR #186, opened 2023-09-28
ci: improve argo templates linting TDE-899

paulfouquet:
#### Motivation
Detecting linting issues before being merged to the master branch.

#### Modification
Creating a test namespace in order to deploy templates and workflows so they can be lint in the current state of the commit.


.github/workflows/main.yml
```diff
@@ -77,6 +77,7 @@ jobs:
           kubectl apply -f config/*.yml --namespace argo
       # Re-lint the workflows after deployment because their dependencies is test with deployement.
       - name: Re-Lint workflows
+        if: github.ref == 'refs/heads/master'
```
blacha, line 80: Would it make sense to

```
Deploy Templates
Lint Workflows
Deploy Workflows
```

I wonder if another option would be to deploy these into a temporary namespace like `argo-topo-workflows-ci` and lint against it there?

paulfouquet: Yes I would totally make sense and I really like your second idea as we can catch those errors before merging to master. And maybe a cleaning step that delete them at the end?

blacha: and possibly a cleaning step at the beginning :D 

I think if you just `k delete namespace some-namespace` it will delete everything inside it too.

amfage (approved):

.github/workflows/main.yml
```diff
@@ -53,9 +53,39 @@ jobs:
...
         run: |
-          ./argo-linux-amd64 lint templates/ -n argo
-          ./argo-linux-amd64 lint workflows/ -n argo
+          # Create test namespace
+          TEST_NAMESPACE=$GITHUB_SHA
+          kubectl create namespace $TEST_NAMESPACE
+
+          # Create copy of Workflows files to change their namespaces
+          mkdir test
+          cp workflows/ test/workflows/ -r
```
l0b0, line 65: Linux tools are by convention called with options *before* files. The same goes for `grep` below.

paulfouquet:
Waiting on the GH action to be uncomment before resolving merging conflicts

.github/workflows/main.yml
```diff
@@ -93,9 +93,39 @@ jobs:
           ./argo-linux-amd64 version
 
       - name: Lint workflows
+        if: github.ref != 'refs/heads/master'
+        continue-on-error: true
```
l0b0, line 97: Do we want [this](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepscontinue-on-error)?

paulfouquet: I wonder if there is another way to delete the namespace (next step) without doing this?

paulfouquet: `if: always()` on the next step should be more appropriate as `continue-on-error` will make the job succeeded

l0b0: [Yep](https://docs.github.com/en/actions/learn-github-actions/expressions#always): `if: ${{ always() }}`

.github/workflows/main.yml
```diff
@@ -122,10 +121,17 @@ jobs:
...
       - name: Delete Test namespace
-        if: github.ref != 'refs/heads/master'
+        if: always()
         run: |
           # Delete the test namespace
-          kubectl delete namespaces "$GITHUB_SHA"
+          work_dir="$(mktemp --directory)"
+          if ! kubectl delete namespaces "$GITHUB_SHA" --dry-run=server 2>"$work_dir"/stderr
+          then 
+              grep 'Error from server (NotFound): namespaces ".*" not found' "$work_dir"/stderr
```
l0b0, line 130: This should be

```
if ! grep …; then
    cat "$work_dir"/stderr >&2
    exit 1
fi
```

My bad.

.github/workflows/main.yml
```diff
@@ -122,10 +121,17 @@ jobs:
...
+          else
+              cat "$work_dir"/stderr >&2
+              exit 1
```
l0b0, line 133: This shouldn't be here, sorry!

.github/workflows/main.yml
```diff
@@ -122,10 +121,17 @@ jobs:
...
+        if: always()
         run: |
           # Delete the test namespace
-          kubectl delete namespaces "$GITHUB_SHA"
+          work_dir="$(mktemp --directory)"
+          if ! kubectl delete namespaces "$GITHUB_SHA" --dry-run=server 2>"$work_dir"/stderr
+          then 
+              grep 'Error from server (NotFound): namespaces ".*" not found' "$work_dir"/stderr
+          else
+              cat "$work_dir"/stderr >&2
```
l0b0, line 132: Instead of this line, we should probably do `2> >(tee "$work_dir"/stderr >&2)` on the `kubectl` line. That way we unconditionally print stderr for debugging purposes (for example, there might be non-fatal warning messages in there).

.github/workflows/main.yml
```diff
@@ -122,10 +121,17 @@ jobs:
           ./argo-linux-amd64 lint test/workflows/ -n "$TEST_NAMESPACE"
 
       - name: Delete Test namespace
-        if: github.ref != 'refs/heads/master'
+        if: always()
         run: |
           # Delete the test namespace
-          kubectl delete namespaces "$GITHUB_SHA"
+          work_dir="$(mktemp --directory)"
```
l0b0, line 127: You could do something like `stderr_file="$(mktemp --directory)/stderr"` to avoid duplication below.

.github/workflows/main.yml
```diff
@@ -93,9 +93,45 @@ jobs:
...
+      - name: Delete Test namespace
+        if: always()
+        run: |
+          # Delete the test namespace
+          stderr_tmp="$(mktemp --directory)/stderr"
+          if ! kubectl delete namespaces fdfgdfgdf --dry-run=server 2> >(tee "$stderr_tmp" >&2)
+          then 
+              if ! grep -q 'Error from server (NotFound): namespaces ".*" not found' "$stderr_tmp"
+              then
+                  exit 1
```
l0b0, line 132: You don't need the inner `if` - if `grep` fails it'll be the exit code of the whole block.

.github/workflows/main.yml
```diff
@@ -93,9 +93,45 @@ jobs:
...
+              sed -i "/^\([[:space:]]*namespace: \).*/s//\1$TEST_NAMESPACE/" "$wf"
+              kubectl apply -f "$wf" --namespace "$TEST_NAMESPACE"
+          done
+
+          # Finally lint the templates
+          ./argo-linux-amd64 lint templates/ -n "$TEST_NAMESPACE"
+          ./argo-linux-amd64 lint test/workflows/ -n "$TEST_NAMESPACE"
+
+      - name: Delete Test namespace
+        if: always()
```
l0b0, line 124: Does this syntax work? The [reference](https://docs.github.com/en/actions/learn-github-actions/expressions#always) uses `if: ${{ always() }}`.

.github/workflows/main.yml
```diff
@@ -93,9 +93,45 @@ jobs:
...
+        run: |
+          # Delete the test namespace
+          stderr_tmp="$(mktemp --directory)/stderr"
+          if ! kubectl delete namespaces fdfgdfgdf --dry-run=server 2> >(tee "$stderr_tmp" >&2)
```
l0b0, line 128: `fdfgdfgdf` should be `"$TEST_NAMESPACE"`

paulfouquet: Oops copied pasted from my test file!

.github/workflows/main.yml
paulfouquet: > Note: The exception to this rule is when you are using expressions in an if clause, where, optionally, you can usually omit ${{ and }}. For more information about if conditionals, see "[Workflow syntax for GitHub Actions](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idif)."

l0b0 (approved):

amfage (approved):

l0b0 (approved):
Lovely, thanks!

PR #187, opened 2023-09-28
fix: publish copy to s3://linz-imagery fails

paulfouquet:


MDavidson17 (approved):

workflows/imagery/publish-odr.yaml
```diff
@@ -123,7 +123,7 @@ spec:
                 - name: version-argo-tasks
                   value: "{{workflow.parameters.version-argo-tasks}}"
                 - name: aws-role-config-path
-                  value: "s3://linz-bucket-config/config-write.open-data-registry.json,s3://linz-bucket-config/config.json"
```
blacha, line 126: How does removing this fix the problem? this should be a priority order?

paulfouquet: This does not fix the problem, it's a separate "fix" that avoids unecessary passing the `config.json` as it's already passed from the `copy.yml` template:
```
env:
          - name: AWS_ROLE_CONFIG_PATH
            value: "s3://linz-bucket-config/config.json,{{inputs.parameters.aws-role-config-path}}"
```

blacha (approved):

PR #193, opened 2023-10-11
feat(cdk8s): use cdk8s for kubernetes config TDE-879

paulfouquet:


.github/workflows/main.yml
```diff
@@ -74,7 +74,10 @@ jobs:
       - name: Deploy config files
         if: github.ref == 'refs/heads/master'
         run: |
-          kubectl apply -f config/*.yml --namespace argo
+          npx cdk8s synth
+          # FIXME: need to change the namespace 
+          kubectl apply -f dist/*.yml --namespace argo
```
blacha, line 79: the namespace should be inside the cdk8s, so shouldnt need it here, and will eventually be across a few namespaces.

blacha (approved):

PR #195, opened 2023-10-15
feat: Fluentbit TDE-913

MDavidson17:
Allows fluentbit deployment via `cdk8s`


config/charts/fluentbit.ts
```diff
@@ -0,0 +1,42 @@
...
+
+import { applyDefaultLabels } from '../util/labels.js';
+
+export interface FluentBitProps {
+  saRoleName: string;
+}
+
+export class FluentBit extends Chart {
+  constructor(scope: Construct, id: string, props: FluentBitProps & ChartProps) {
+    super(scope, id, applyDefaultLabels(props, 'aws-for-fluent-bit', '2.31.11', 'logs', 'workflows'));
```
paulfouquet, line 12: It's a bit annoying to maintain both chart and app version when upgrading because they are different and you need to look up at different places to get them

config/charts/fluentbit.ts
```diff
@@ -0,0 +1,42 @@
...
+        Format regex
+        Regex ^(?<time>[^ ]+) (?<stream>stdout|stderr) (?<logtag>[^ ]*) (?<log>.*)$
+        Time_Key    time
+        Time_Format %Y-%m-%dT%H:%M:%S.%L%z
+    `;
+
+    new Helm(this, 'aws-for-fluent-bit', {
+      chart: 'aws-for-fluent-bit',
+      repo: 'https://aws.github.io/eks-charts',
+      namespace: 'fluent-bit',
```
paulfouquet, line 28: should we name it `fluentbit` for easiness?

paulfouquet (approved):

config/cdk8s.ts
```diff
@@ -1,35 +1,42 @@
 import { App } from 'cdk8s';
 
 import { ArgoSemaphore } from './charts/argo.semaphores';
+import { FluentBit } from './charts/fluentbit';
 import { Karpenter, KarpenterProvisioner } from './charts/karpenter';
 import { CoreDns } from './charts/kube-system.coredns';
-import { CfnOutputKeys } from './constants';
-import { CLUSTER_NAME } from './constants';
+import { CfnOutputKeys, ClusterName } from './constants';
```
l0b0, line 7: Would it be possible to split this sort of refactoring into a separate PR? It would make the whole much easier to review.

config/charts/fluentbit.ts
```diff
@@ -0,0 +1,65 @@
...
+  constructor(scope: Construct, id: string, props: FluentBitProps & ChartProps) {
+    super(scope, id, applyDefaultLabels(props, 'aws-for-fluent-bit', awsForFluentBitVersion, 'logs', 'workflows'));
+
+    const FluentParserName = 'containerd';
+    // This needs to be properly formatted, and it was taken directly from https://github.com/microsoft/fluentbit-containerd-cri-o-json-log
+    // The key part is the message must be parsed as "log" otherwise it wont be parsed as JSON
+    const extraParsers = `[PARSER]
+    Name ${FluentParserName}
+    Format regex
+    Regex ^(?<time>[^ ]+) (?<stream>stdout|stderr) (?<logtag>[^ ]*) (?<log>.*)$
```
l0b0, line 23: If I understand correctly `(?<time>[^ ]+)` would just match a sequence of one or more "^" or space characters. That doesn't seem right.

blacha: This regex was stolen from the internet (months ago) as a example on how to parse docker logs, and then copied from another repo to here.

config/cdk8s.ts
blacha: It is generally a good idea to do this, or even as separate commit so it can be viewed separately.

This change was done in a peer programming session as a way to fix a import error.

config/charts/fluentbit.ts
```diff
@@ -0,0 +1,65 @@
...
+    Time_Key    time
+    Time_Format %Y-%m-%dT%H:%M:%S.%L%z
+    `;
+
+    const extraService = `
+HTTP_Server  Off
```
blacha, line 29: Would be great to doc why we are turning the HTTP server off

config/charts/fluentbit.ts
```diff
@@ -0,0 +1,65 @@
...
+          logRetentionDays: 30,
+          logGroupName: `/aws/eks/${props.clusterName}/logs`,
+          logGroupTemplate: `/aws/eks/${props.clusterName}/workload/$kubernetes['namespace_name']`,
+          logStreamPrefix: 'fb-',
+        },
+        firehose: { enabled: false },
+        kinesis: { enabled: false },
+        elasticsearch: { enabled: false },
+        service: { extraParsers, extraService },
+        livenessProbe: false,
```
blacha, line 60: And doc why these are both off aswell.

config/charts/karpenter.ts
```diff
@@ -69,6 +69,7 @@ export class Karpenter extends Chart {
       namespace: 'karpenter',
       version: 'v0.31.1',
       values: {
+        // FIXME: fullnameOverride: 'karpenter',
```
blacha, line 72: Could this be expanded, I am not sure what this fixme refers to ?

config/charts/fluentbit.ts
l0b0: My bad, I momentarily forgot how regexes work! `[^ ]` matches *every character except for space.*

paulfouquet (approved):

amfage (approved):

PR #196, opened 2023-10-17
feat: config karpenter TDE-903

paulfouquet:


.gitignore
```diff
@@ -2,3 +2,4 @@ node_modules/
 dist/
 cdk.out/
 cdk.context.json
+config/imports/
```
blacha, line 5: We need to commit these files (?)

paulfouquet: You are right, it is even [_recommended_ ](https://cdk8s.io/docs/latest/cli/import/#kubernetes-apis)

blacha (approved):

PR #203, opened 2023-10-20
docs: update readme for connecting to new cluster

blacha:


paulfouquet (approved):

paulfouquet (commented):
Wonder if we should mention `kubectl` instead of `k` to make it clearer (if people haven't set up any alias)

blacha:
> Wonder if we should mention `kubectl` instead of `k` to make it clearer (if people haven't set up any alias)

a few lines above it says to use k instead of kubectl to make your life easier.

paulfouquet:
> > Wonder if we should mention `kubectl` instead of `k` to make it clearer (if people haven't set up any alias)
> 
> a few lines above it says to use k instead of kubectl to make your life easier.

Oh I haven't scrolled up enough! sorry

PR #205, opened 2023-10-25
refactor: link both provisioners back to a shared node template

blacha:
#### Motivation

There appears to be a bug with cdk8s unsing nested specs causing volumes to show as 

```
volumeSize:
  value: 200Gi
```

the value key is invalid, doing the exact same source as a node template outpus

```
volumeSize: 200Gi
```

to work around the problem use a shared node template.

#### Modification

Adds a shared aws node template for both spot provisionsers to reference.

#### Checklist

- [ ] Tests updated
- [ ] Docs updated
- [ ] Issue linked in Title

### DIFF
--- a/config/charts/karpenter.ts
+++ b/config/charts/karpenter.ts
@@ -109,7 +109,7 @@ export class KarpenterProvisioner extends Chart {
     });
 
     const provisionAmd64 = new Provisioner(this, 'ClusterAmd64WorkerNodes', {
-      metadata: { name: `karpenter-amd64`.toLowerCase(), namespace: 'karpenter' },
+      metadata: { name: `karpenter-amd64-spot`, namespace: 'karpenter' },
       spec: {
         // Ensure only pods that tolerate spot run on spot instance types
         // to prevent long running pods (eg kube-dns) being moved.
@@ -126,7 +126,7 @@ export class KarpenterProvisioner extends Chart {
     });
 
     const provisionArm64 = new Provisioner(this, 'ClusterArmWorkerNodes', {
-      metadata: { name: `karpenter-arm64`.toLowerCase(), namespace: 'karpenter' },
+      metadata: { name: `karpenter-arm64-spot`, namespace: 'karpenter' },
       spec: {
         taints: [
           // Instances that want ARM have to tolerate the arm taint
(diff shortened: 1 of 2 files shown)


config/charts/karpenter.ts
```diff
@@ -39,13 +64,13 @@ export class Karpenter extends Chart {
...
-    new Helm(this, 'karpenter', {
+    const karpenter = new Helm(this, 'karpenter', {
       chart: 'oci://public.ecr.aws/karpenter/karpenter',
       namespace: 'karpenter',
-      version: 'v0.31.0',
+      version: 'v0.31.1',
       values: {
         serviceAccount: {
-          create: false,
+          create: true,
```
blacha, line 73:
```suggestion
          create: false,
```

config/charts/karpenter.ts
```diff
@@ -97,14 +119,14 @@ export class KarpenterProvisioner extends Chart {
           { key: 'kubernetes.io/arch', operator: 'In', values: ['amd64'] },
           { key: 'karpenter.k8s.aws/instance-family', operator: 'In', values: ['c5', 'c6i', 'c6a'] },
         ],
-        limits: { resources: { cpu: ProvisionerSpecLimitsResources.fromString('20000m') } },
-        provider,
+        limits: { resources: { cpu: ProvisionerSpecLimitsResources.fromNumber(2000) } },
```
l0b0, line 129: Is this change intentional?

config/charts/karpenter.ts
```diff
@@ -18,10 +43,10 @@ export class Karpenter extends Chart {
     super(scope, id, applyDefaultLabels(props, 'karpenter', 'v0.31.0', 'karpenter', 'workflows'));
 
     // Deploying the CRD
-    new Helm(this, 'karpenter-crd', {
+    const crd = new Helm(this, 'karpenter-crd', {
       chart: 'oci://public.ecr.aws/karpenter/karpenter-crd',
       namespace: 'karpenter',
-      version: 'v0.31.0',
+      version: 'v0.31.1',
```
blacha, line 49: This should be a constant

config/charts/karpenter.ts
blacha: Yup, the cluster needs more than 20 cpus, this allows 2000 cpus to be spun up. (also likely too small)

paulfouquet (approved):

PR #206, opened 2023-10-25
fix: use template rcode NOERROR to prevent timeouts

blacha:
#### Motivation

When using `rewrite stop type AAAA A` DNS resolution would often hang for 30+seconds on the clients waiting for a response

#### Modification

Instantly respond to AAAA requests with a empty NOERROR to force clients to use A records for external hosts.

#### Checklist

- [ ] Tests updated
- [ ] Docs updated
- [ ] Issue linked in Title

### DIFF
--- a/cdk8s.yaml
+++ b/cdk8s.yaml
@@ -1,4 +1,4 @@
-app: npx tsx config/cdk8s.ts
+app: npx tsx config/cdk8s.old.cluster.ts
 language: typescript
 imports:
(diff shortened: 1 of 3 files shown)


cdk8s.yaml
```diff
@@ -1,4 +1,4 @@
-app: npx tsx config/cdk8s.ts
+app: npx tsx config/cdk8s.old.cluster.ts
```
paulfouquet, line 1:
```suggestion
app: npx tsx config/cdk8s.ts
```

paulfouquet (approved):

MDavidson17 (approved):

PR #214, opened 2023-10-29
feat: example of running a workflow on spot arm nodes

blacha:
#### Motivation

I can never remember what the selectors and tolerations are to run ARM based workflows are so create a example workflow that runs on ARM SPOT

#### Modification

Example workflow to run on SPOT ARM nodes.

#### Checklist

- [ ] Tests updated
- [ ] Docs updated
- [ ] Issue linked in Title


l0b0 (approved):
LGTM, although I don't know Argo itself.

PR #215, opened 2023-10-29
fix: correct cloudflare namespace & container

blacha:
#### Motivation

cloudflared was being deployed into the wrong namespace and using a invalid container link

#### Modification

corrects the namespace and container url.

#### Checklist

- [ ] Tests updated
- [ ] Docs updated
- [ ] Issue linked in Title

### DIFF
--- a/infra/charts/cloudflared.ts
+++ b/infra/charts/cloudflared.ts
@@ -46,7 +46,7 @@ export class Cloudflared extends Chart {
       containers: [
         {
           name: 'cloudflared',
-          image: props.accountId + '.dkr.ecr.ap-southeast-2.amazonaws.com/eks:cloudflared-2023.8.2',
+          image: '019359803926.dkr.ecr.ap-southeast-2.amazonaws.com/cloudflared:2023.8.2',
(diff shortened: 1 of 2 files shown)


infra/charts/cloudflared.ts
```diff
@@ -46,7 +46,7 @@ export class Cloudflared extends Chart {
       containers: [
         {
           name: 'cloudflared',
-          image: props.accountId + '.dkr.ecr.ap-southeast-2.amazonaws.com/eks:cloudflared-2023.8.2',
+          image: '019359803926.dkr.ecr.ap-southeast-2.amazonaws.com/cloudflared:2023.8.2',
```
l0b0, line 49: Why hard-code the account ID? Is it a third party account?

blacha: Docker rate limits us so we cannot pull directly from the docker registry, so we mirror this container internally for all of LINZ.

`props.accountId` is a kubernetes id and has nothing to do with EKS so the account id needs to come from somewhere, all throughout this repo is this account id hard coded.

l0b0 (approved):

PR #223, opened 2023-11-02
feat: Argo Workflows archive feature using RDS TDE-883 TDE-731

amfage:
This is a Proof of Concept to add the configuration to connect Argo to RDS and enable the Workflow Archive feature.
A decision still needs to be made about the best way to provision secrets into EKS.
There are also decisions to make about how the database and workflow archive feature should be configured.
It also needs refactoring to validate cfnOutput keys from multiple stacks.


amfage:
Superseded https://github.com/linz/topo-workflows/pull/284

PR #226, opened 2023-11-05
feat: add argo schema url for all argo docs

blacha:
#### Motivation

Now that there are multiple formats of  yaml documents in this repository it is helpful to be specific for IDEs to know what sort of yaml document each are (k8s vs ArgoWorkflows.

#### Modification

Forces IDE language servers to use a consistent version of the Argo Workflows schemas

#### Checklist

- [ ] Tests updated
- [ ] Docs updated
- [ ] Issue linked in Title

### DIFF
--- a/templates/argo-tasks/copy.yml
+++ b/templates/argo-tasks/copy.yml
@@ -1,3 +1,5 @@
+# yaml-language-server: $schema=https://raw.githubusercontent.com/argoproj/argo-workflows/v3.4.13/api/jsonschema/schema.json
+
 apiVersion: argoproj.io/v1alpha1
 kind: WorkflowTemplate
 metadata:
@@ -8,6 +10,7 @@ spec:
   templateDefaults:
     container:
       imagePullPolicy: Always
+      image: ''
   entrypoint: main
   templates:
     - name: main
--- a/workflows/test/flatten.yaml
+++ b/workflows/test/flatten.yaml
@@ -1,4 +1,5 @@
----
+# yaml-language-server: $schema=https://raw.githubusercontent.com/argoproj/argo-workflows/v3.4.13/api/jsonschema/schema.json
+
 apiVersion: argoproj.io/v1alpha1
 kind: WorkflowTemplate
 metadata:
@@ -25,6 +26,7 @@ spec:
   templateDefaults:
     container:
       imagePullPolicy: Always
+      iamge: ''
   templates:
     - name: main
       dag:
--- a/workflows/test/sleep.yml
+++ b/workflows/test/sleep.yml
@@ -1,3 +1,5 @@
+# yaml-language-server: $schema=https://raw.githubusercontent.com/argoproj/argo-workflows/v3.4.13/api/jsonschema/schema.json
+
 apiVersion: argoproj.io/v1alpha1
 kind: Workflow
 metadata:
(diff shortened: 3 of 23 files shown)


workflows/test/sleep.yml
```diff
@@ -13,5 +15,5 @@ spec:
           requests:
             memory: 3.9Gi
             cpu: 2000m
-        image: ubuntu:22.04
+        image: 019359803926.dkr.ecr.ap-southeast-2.amazonaws.com/argo-tasks:latest
```
blacha, line 18: We consistently get rate limited by docker hub and am often changing this, so it accidentally snuck it but :shrug: is a good change to have :P

templates/argo-tasks/copy.yml
```diff
@@ -1,3 +1,5 @@
+# yaml-language-server: $schema=https://raw.githubusercontent.com/argoproj/argo-workflows/v3.4.13/api/jsonschema/schema.json
```
l0b0, line 1: Is v3.4.13 specifically the one we should be using?

blacha: it should be aligned to the version in cdk8s, but we haven't deployed it yet, and I dont see a easy way of referencing it back to the cdk8s version.

l0b0: I guess we could just add a comment to the cdk8s dependency definition in `package.json` to let people know it should be kept in sync?

workflows/test/flatten.yaml
```diff
@@ -25,6 +26,7 @@ spec:
   templateDefaults:
     container:
       imagePullPolicy: Always
+      iamge: ''
```
l0b0, line 29: Typo here.

templates/argo-tasks/copy.yml
paulfouquet: I think Blayne is refering to the Argo Workflows version https://github.com/linz/topo-workflows/blob/09976f2c628136be533346de98bfc25c64d6a87a/infra/charts/argo.workflows.ts#L39, not the cdk8s dependency version

l0b0: Oh right, thank you! That explains why I couldn't find "v3.4.13" anywhere else in the code base. Should we not use that version?

paulfouquet: Ideally should match the production cluster Argo version. The production has not been updated as it's tricky to update it (that's why we are doing this current work with  CDK8S). We'll update it and make sure both version are matching when the new cluster will be deployed as the production one.

l0b0: Aren't you going to run into problems trying to deploy with mismatching versions? That is, `const appVersion = 'A';` and `# yaml-language-server: $schema=https://raw.githubusercontent.com/argoproj/argo-workflows/B/api/jsonschema/schema.json`.

blacha: 3.4.13 was released today, so I was going to do another PR to get argo to 3.4.13, at that time I can add some docs about keeping the versions in sync.

If we wanted to go all our we could get the cdk8s to download the specific schema doc and then we could just use `../../argo-schema.json` 

I dont think the versions change that often and hopefully a patch version missmatch wouldnt have conflicts I could see lots of issues if we were doing a 3.x.x to 4.x.x schema validation though.

paulfouquet (approved):

PR #228, opened 2023-11-07
fix: labels cannot contain "/"

blacha:
#### Motivation

Deployments are broken as they use a label with a "/", labels cannot contain "/" in their value

> Service "fluentbit" is invalid: metadata.labels: Invalid value: "linz/topo-workflows": a valid label must be an empty string or consist of alphanumeric characters, '-', '_' or '.', and must start and end with an alphanumeric character (e.g. 'MyValue',  or 'my_value',  or '12345', regex used for validation is '(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])?')

#### Modification

Replaces the "/" in the label with "__"

#### Checklist

- [ ] Tests updated
- [ ] Docs updated
- [ ] Issue linked in Title

### DIFF
--- a/infra/util/labels.ts
+++ b/infra/util/labels.ts
@@ -22,7 +22,8 @@ export function defaultLabels(
     'app.kubernetes.io/part-of': partOf,
     'app.kubernetes.io/managed-by': 'cdk8s',
     // TODO these are not "well-known" maybe we should move them into the linz namespace?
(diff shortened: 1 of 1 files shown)


paulfouquet (approved):

l0b0 (approved):

l0b0:
This is what we get from the "whitelist rather than escape" school :sigh:.

PR #229, opened 2023-11-07
fix: fluentbit is not deployed on new nodes

paulfouquet:
#### Motivation

Fluentbit is not deployed on new nodes because of the karpenter provisioner taints.

#### Modification

Add a toleration to the daemon set matching the provisioner taints.

#### Checklist

- [ ] Tests updated
- [ ] Docs updated
- [ ] Issue linked in Title

### DIFF
--- a/infra/charts/fluentbit.ts
+++ b/infra/charts/fluentbit.ts
@@ -88,8 +88,8 @@ HC_Period 5
         livenessProbe: false,
         readinessProbe: false,
         tolerations: [
-          { key: 'karpenter.sh/capacity-type', value: 'spot', effect: 'NoSchedule', operator: 'Equal' },
-          { key: 'kubernetes.io/arch', value: 'arm64', effect: 'NoSchedule', operator: 'Equal' },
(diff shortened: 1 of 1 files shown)


infra/charts/fluentbit.ts
```diff
@@ -87,6 +87,10 @@ HC_Period 5
         // FIXME: `livenessProbe` and `readinessProbe` deactivated https://github.com/aws/eks-charts/issues/995
         livenessProbe: false,
         readinessProbe: false,
+        tolerations: [
+          { key: 'karpenter.sh/capacity-type', value: 'spot', effect: 'NoSchedule', operator: 'Equal' },
```
l0b0, line 91: Can I read this like "if the key 'karpenter.sh/capacity-type' has value Equal to 'spot' then 'NoSchedule'"? If so, would it make sense to put the keys in that order? Basically, `{ key: 'karpenter.sh/capacity-type', operator: 'Equal', value: 'spot', effect: 'NoSchedule' },`

l0b0 (approved):

blacha (approved):

PR #244, opened 2023-11-08
feat: lint inputs before publish TDE-857

MDavidson17:
#### Motivation

To be sure that the target paths in the publish workflows follow conventions by linting.

#### Modification

new linting step added to the start of `publish-copy` and `publish-odr`
This step can be skipped with the new parameter `skip-lint: 'true'`

#### Checklist

~~- [ ] Tests updated~~ - there are no tests but the template was tested manually
- [x] Docs updated
- [x] Issue linked in Title

### DIFF
--- a/templates/argo-tasks/README.md
+++ b/templates/argo-tasks/README.md
@@ -1,5 +1,13 @@
 # ArgoTasks Templates
 
+## Contents:
+
+- [Group](##argo-tasks/group)
+- [Copy](##argo-tasks/copy)
+- [Create Manifest](##argo-tasks/create-manifest)
+- [Push to Github](##argo-tasks/push-to-github)
+- [Lint Inputs](##argo-tasks/lint-inputs)
+
 ## argo-tasks/group - `tpl-at-group`
 
 Group inputs into outputs to be used with `withParam` to run one task per grouping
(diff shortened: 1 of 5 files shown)


workflows/imagery/README.md
```diff
@@ -201,6 +201,7 @@ Access permissions are controlled by the [Bucket Sharing Config](https://github.
 | group       | int   | 1000                                          | The maximum number of files for each pod to copy (will use the value of `group` or `group-size` that is reached first).                                          |
 | group-size  | str   | 100Gi                                         | The maximum group size of files for each pod to copy (will use the value of `group` or `group-size` that is reached first).                                      |
 | transform   | str   | `f`                                           | String to be transformed from source to target to renamed filenames, e.g. `f.replace("text to replace", "new_text_to_use")`. Leave as `f` for no transformation. |
+| skip-lint   | str   | `false`                                       | False: lint the target path; True: skip linting of target path - to be used when publishing to a location other than `linz-imagery` or using a non-standard path |
```
l0b0, line 204: How about calling this `lint` to shorten and avoid a semantic double negative? "Skip linting" isn't precisely a negative, but in this particular case it could be equivalent to saying "Don't lint". Also, flipping it should make the conditionals elsewhere simpler.

templates/argo-tasks/README.md
```diff
@@ -160,3 +168,24 @@ See https://github.com/linz/argo-tasks#stac-github-import
...
+  templateRef:
+    name: tpl-at-lint-inputs
+    template: main
+  arguments:
+    parameters:
+      - name: version
+        value: '{{workflow.parameters.version-argo-tasks}}'
+      - name: path
+        value: '{{workflow.parameters.target}}'
+  when: '{{workflow.parameters.lint}} == true'
```
l0b0, line 190: I think we can shorten this to `when: '{{workflow.parameters.lint}}'`. Ditto for the other identical comparisons.

templates/argo-tasks/lint-inputs.yml
```diff
@@ -0,0 +1,31 @@
+apiVersion: argoproj.io/v1alpha1
```
paulfouquet, line 1:
```suggestion
# yaml-language-server: $schema=https://raw.githubusercontent.com/argoproj/argo-workflows/v3.4.13/api/jsonschema/schema.json

apiVersion: argoproj.io/v1alpha1
```

l0b0 (approved):

PR #251, opened 2023-11-09
ci: store cluster name as env variable for re-use

paulfouquet:
#### Motivation

I like using constants so I don't have to use find and replace and make mistakes.

#### Modification

Store the cluster name as a environment variable.

#### Checklist

- [ ] Tests updated - no tests for GHA
- [ ] Docs updated - no docs necessary for this change
- [ ] Issue linked in Title - do not need a ticket for this kind of change

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -4,12 +4,12 @@ jobs:
   main:
     name: Format & Deploy(master)
     runs-on: ubuntu-latest
-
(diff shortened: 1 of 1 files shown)


blacha:
This is going to conflict with #246 could we merge the two together?

l0b0 (dismissed):

l0b0 (approved):

PR #278, opened 2023-11-15
feat: give maintainers access to the EKS cluster TDE-926

paulfouquet:
#### Motivation

Maintainers should have direct access to the EKS Cluster.

#### Modification

Pass the maintainer role ARNs as a CDK context argument so they can be added as master on the cluster

#### Checklist

- [x] Tests updated
- [ ] Docs updated
- [x] Issue linked in Title

### DIFF
--- a/infra/eks/arn.ts
+++ b/infra/eks/arn.ts
@@ -42,8 +42,11 @@ export function tryGetContextArn(node: Node, context: string): string | null {
  * @returns arns if they are valid, null otherwise
  */
 export function tryGetContextArns(node: Node, context: string): string[] | null {
-  const ctx = node.tryGetContext(context);
+  let ctx = node.tryGetContext(context);
   if (ctx == null) return null;
+  if (typeof ctx === 'string') {
+    ctx = ctx.split(',');
+  }
   if (!Array.isArray(ctx)) throw new Error('Failed to parse ARN, is not a string[]');
   for (const arn of ctx) validateRoleArn(arn);
   return ctx;
--- a/infra/eks/cluster.ts
+++ b/infra/eks/cluster.ts
@@ -17,8 +17,8 @@ import { Construct } from 'constructs';
 import { CfnOutputKeys, ScratchBucketName } from '../constants.js';
 
 interface EksClusterProps extends StackProps {
-  /** CI User to grant access to the cluster */
-  ciRoleArn: string;
+  /** List of role ARNs to grant access to the cluster */
+  maintainerRoleArns: string[];
 }
 
 export class LinzEksCluster extends Stack {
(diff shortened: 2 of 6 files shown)


package.json
```diff
@@ -18,7 +18,8 @@
   },
   "scripts": {
     "lint": "npx eslint . --ignore-path .gitignore",
-    "format": "npx prettier . -w"
+    "format": "npx prettier . -w",
+    "test": "node --loader tsx --test infra/**/*.test.ts"
```
blacha, line 22: does this need to be restricted to `infra/` does it work without it?

infra/eks/cluster.ts
```diff
@@ -71,14 +72,11 @@ export class LinzEksCluster extends Stack {
...
-    this.cluster.awsAuth.addMastersRole(accountAdminRole);
-
-    // If defined allow the CI user access to the cluster
-    if (props.ciRoleArn) {
-      const ciRole = Role.fromRoleArn(this, 'CiRole', props.ciRoleArn);
-      this.cluster.awsAuth.addMastersRole(ciRole);
+    //allow the maintainer roles access to the cluster
+    for (const roleArn of props.maintainerRoleArns) {
+      const roleId = createHash('sha256').update(roleArn).digest('hex').slice(0, 12);
+      const role = Role.fromRoleArn(this, roleId, roleArn);
```
blacha, line 78: should have some sort of prefix about the id, and needs a default policy doc

`AdminRole-${roleId}` ?

infra/eks/arn.ts
```diff
@@ -42,8 +42,11 @@ export function tryGetContextArn(node: Node, context: string): string | null {
  * @returns arns if they are valid, null otherwise
  */
 export function tryGetContextArns(node: Node, context: string): string[] | null {
-  const ctx = node.tryGetContext(context);
+  let ctx = node.tryGetContext(context);
   if (ctx == null) return null;
```
l0b0, line 46: We could possibly return `[]` in this case, to avoid extra type handling.

package.json
paulfouquet: I can't make it work without specifying `infra`

blacha: what about `... --test infra/` I think it should be looking for all `.test.ts` maybe its not looking for ts files.

blacha: hmm maybe not, I think ** doesnt work in windows but :shrug: we don't build in windows.

infra/eks/arn.ts
blacha: null is more common, as you would still need to do `if (ret.length === 0) throw error`

Using `?` means it you can generally ignore the fact that it returned null eg

```typescript
const arns = tryGetContextArns(...)
const arn = arns?.find(f => f.includes("internal-user"))
if (arn == null) throw new Error("Cannot find internal-user role")
```

package.json
l0b0: Probably `**` could be just `*` - most contexts don't support recursive globs.

paulfouquet: I can change it to `*` if needed

blacha: I would leave as is, as `*` wouldn't cover `infra/foo/bar/test.ts` a more complex option would be to use `find` or a third party glob eg `glob -c ".." "**/.test.ts"`

l0b0 (approved):

blacha (approved):

PR #284, opened 2023-11-16
feat: Argo Workflows archive feature using RDS TDE-731

amfage:
#### Motivation

The aim of this change is to enable the Archived Workflows in Argo so that older workflow runs can be referenced in the future e.g. to see what was run when, what parameters were used.

#### Modification

This change provisions an AWS RDS PostgreSQL database and configures Argo Workflows to use it, so that the Archived Workflows feature is enabled.

#### Checklist

- [N/A] Tests updated
- [x] Docs updated
- [x] Issue linked in Title

### DIFF
--- a/infra/cdk8s.ts
+++ b/infra/cdk8s.ts
@@ -27,6 +27,9 @@ async function main(): Promise<void> {
 
     // Personal access token to gain access to linz-basemaps github user
     githubPat: '/eks/github/linz-basemaps/pat',
+
+    // Argo Database connection password
+    argoDbPassword: '/eks/argo/postgres/password',
   });
 
   const coredns = new CoreDns(app, 'dns', {});
@@ -59,6 +62,8 @@ async function main(): Promise<void> {
     clusterName: ClusterName,
     saName: cfnOutputs[CfnOutputKeys.ArgoRunnerServiceAccountName],
     tempBucketName: ScratchBucketName,
+    argoDbEndpoint: cfnOutputs[CfnOutputKeys.ArgoDbEndpoint],
+    argoDbPassword: ssmConfig.argoDbPassword,
   });
 
   new ArgoExtras(app, 'argo-extras', {
--- a/infra/charts/argo.workflows.ts
+++ b/infra/charts/argo.workflows.ts
@@ -2,8 +2,7 @@ import { Chart, ChartProps, Duration, Helm } from 'cdk8s';
 import { Secret } from 'cdk8s-plus-27';
 import { Construct } from 'constructs';
 
-import { ArgoDbUser, CfnOutputKeys, ClusterName, validateKeys } from '../constants.js';
-import { getCfnOutputs } from '../util/cloud.formation.js';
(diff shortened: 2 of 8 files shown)


infra/charts/argo.workflows.ts
```diff
@@ -22,6 +24,13 @@ export interface ArgoWorkflowsProps {
    * @example "Workflows"
    */
   clusterName: string;
+  /**
+   * The Argo database endpoint
+   *
+   * @example "argodb-argodb4be14fa2-p8yjinijwbro.cmpyjhgv78aj.ap-southeast-2.rds.amazonaws.com"
+   */
+  argoDbEndpoint: string;
+  argoDbPassword: string;
```
blacha, line 33: could doc this, it is pretty self explanatory

infra/constants.ts
```diff
@@ -2,11 +2,19 @@
 export const ClusterName = 'Workflows';
 /* LINZ conventional name for Argo Workflows artifact bucket */
 export const ScratchBucketName = `linz-${ClusterName.toLowerCase()}-scratch`;
+/* Argo Database Instance name */
+export const ArgoDbInstanceName = 'ArgoDb';
+/* Argo Database user */
```
blacha, line 7: docs?

blacha (approved):

infra/charts/argo.workflows.ts
```diff
@@ -22,6 +24,18 @@ export interface ArgoWorkflowsProps {
...
+   *
+   * @example "argodb-argodb4be14fa2-p8yjinijwbro.cmpyjhgv78aj.ap-southeast-2.rds.amazonaws.com"
+   */
+  argoDbEndpoint: string;
+  /**
+   * The Argo database password
+   *
+   * @example "eighoo5room0aeM^ahz0Otoh4aakiipo"
+   */
+  argoDbPassword: string;
```
paulfouquet, line 38: Where does it come from?

amfage: The parameter store.

paulfouquet: Yes but how does it arrive to the parameter store?

blacha: The same as all the other secrets? manually

paulfouquet: I guess my question is more around where do you get the password from, did you need to set up something in the AWS side

blacha: Yup, we should manually generate a longish password to use, We should (in a later PR) document how to rotate the password.

infra/cdk8s.ts
```diff
@@ -27,6 +27,9 @@ async function main(): Promise<void> {
 
     // Personal access token to gain access to linz-basemaps github user
     githubPat: '/eks/github/linz-basemaps/pat',
+
+    // Argo Database connection password
+    argoDbPassword: '/eks/argo/postgres/password',
```
paulfouquet, line 32: That would be useful to store those secret path as constant. The same path is used in two places

paulfouquet (approved):

PR #298, opened 2023-11-21
fix: imagery-import-cogify use Basemaps PAT TDE-953

amfage:
#### Motivation

The new cluster does not contain all the old github ssh keys so the Basemaps workflows that are using them need to move to the Basemaps PAT which does exist on the new cluster.

#### Modification

Remove the old secret reference in imagery-import-cogify and use the Basemaps PAT instead.
Remove the imagery-import workflow that was using the old github ssh key as this workflow has been superseded by imagery-import-cogify.
Update the standardising-publish-import workflow that was still using the old imagery-import workflow.

#### Checklist

- [ ] Tests updated
- [ ] Docs updated
- [X] Issue linked in Title


workflows/imagery/standardising-publish-import.yaml
```diff
@@ -41,8 +41,12 @@ spec:
         value: 'true'
       - name: retile
         value: 'false'
+      - name: version_argo_tasks
```
blacha, line 44: we now have two versions for argo tasks?

amfage: One has underscores and the other dashes! It is not ideal. Were we going to move everything to underscores at some point?

blacha: underscores would be best, saves us doing this `['foo-bar']` vs `.foo_bar`

amfage: Maybe not a change for this PR, though.

blacha: I don't seem them used in this file though?

amfage: It's for the downstream WorkflowTemplate `basemaps-imagery-import-cogify`

amfage:
Out of date, will replace when `github-linz-basemaps-pat` is renamed to `github-linz-li-bot-pat`

PR #307, opened 2023-11-27
feat: add region label to imagery related workflows TDE-957, TDE-955

paulfouquet:
#### Motivation

The region of the dataset is a useful information to be able to filter on.

#### Modification

`region` is passed as a parameter of the workflow to the `linz.govt.nz/region` label value. This is a mandatory field proposed as an enum to the user. Any modification of the `region` list will have to be applied at all the places it is used.
This change is also part of TDE-955 that will use this workflow parameter as an input of a task/step to build the dataset title.

#### Checklist

- [ ] Tests updated - N/A
- [x] Docs updated
- [x] Issue linked in Title

### DIFF
--- a/workflows/basemaps/imagery-import-cogify.yml
+++ b/workflows/basemaps/imagery-import-cogify.yml
@@ -21,6 +21,8 @@ spec:
     labelsFrom:
       linz.govt.nz/ticket:
         expression: workflow.parameters.ticket
+      linz.govt.nz/region:
+        expression: workflow.parameters.region
   arguments:
     parameters:
       - name: version_basemaps_cli
@@ -34,6 +36,28 @@ spec:
       - name: ticket
         description: Ticket ID e.g. 'AIP-55'
         value: ''
+
+      - name: region
+        description: Region of the dataset
+        value: 'new-zealand'
+        enum:
+          - 'auckland'
+          - 'bay-of-plenty'
+          - 'canterbury'
+          - 'gisborne'
+          - 'hawkes-bay'
+          - 'manawatu-whanganui'
+          - 'marlborough'
+          - 'nelson'
+          - 'new-zealand'
+          - 'northland'
+          - 'otago'
+          - 'southland'
+          - 'taranaki'
+          - 'tasman'
+          - 'waikato'
+          - 'wellington'
+          - 'west-coast'
 
       - name: source
         description: Source imagery location "s3://linz-imagery"
--- a/workflows/basemaps/imagery-import-cogify.yml
+++ b/workflows/basemaps/imagery-import-cogify.yml
@@ -41,7 +41,7 @@ spec:
         description: Region of the dataset
         value: 'new-zealand'
         enum:
-          - 'antartica'
+          - 'antarctica'
           - 'auckland'
           - 'bay-of-plenty'
           - 'canterbury'
--- a/workflows/basemaps/imagery-import-cogify.yml
+++ b/workflows/basemaps/imagery-import-cogify.yml
@@ -41,10 +41,12 @@ spec:
         description: Region of the dataset
         value: 'new-zealand'
         enum:
+          - 'antartica'
           - 'auckland'
           - 'bay-of-plenty'
           - 'canterbury'
           - 'gisborne'
+          - 'global'
           - 'hawkes-bay'
           - 'manawatu-whanganui'
           - 'marlborough'
@@ -52,6 +54,7 @@ spec:
           - 'new-zealand'
           - 'northland'
           - 'otago'
+          - 'pacific-islands'
           - 'southland'
           - 'taranaki'
           - 'tasman'
--- a/workflows/basemaps/imagery-import.yaml
+++ b/workflows/basemaps/imagery-import.yaml
@@ -27,7 +27,7 @@ spec:
         description: Region of the dataset
         value: 'new-zealand'
         enum:
-          - 'antartica'
+          - 'antarctica'
           - 'auckland'
           - 'bay-of-plenty'
           - 'canterbury'
--- a/workflows/basemaps/imagery-import.yaml
+++ b/workflows/basemaps/imagery-import.yaml
@@ -27,10 +27,12 @@ spec:
         description: Region of the dataset
         value: 'new-zealand'
         enum:
+          - 'antartica'
           - 'auckland'
           - 'bay-of-plenty'
           - 'canterbury'
           - 'gisborne'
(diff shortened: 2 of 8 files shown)


workflows/basemaps/imagery-import-cogify.yml
```diff
@@ -35,6 +37,28 @@ spec:
         description: Ticket ID e.g. 'AIP-55'
         value: ''
 
+      - name: region
+        description: Region of the dataset
+        value: 'new-zealand'
+        enum:
```
blacha, line 43: do we need a other/na for things that our outside this scope?

paulfouquet: I would have thought `new-zealand` would be the one to use for now. Have we processed something outside of NZ yet?

blacha: Gebco?

https://www.gebco.net/data_and_products/gridded_bathymetry_data/

paulfouquet: I'm not against adding a `other/na`, but maybe we would like to always be specific on what we process and add a new value when we know what it is. On the other side that can be annoying to be blocked by having to add the new value before processing stuffs...

dwsilk: add `global`, `pacific-islands` and `antarctica` to each enum
that will cover everything I can think of at the moment

workflows/basemaps/imagery-import-cogify.yml
```diff
@@ -35,6 +37,31 @@ spec:
...
+          - 'antartica'
```
amfage, line 44: `antarctica`
(also in the other workflows)

workflows/raster/README.md
```diff
@@ -256,6 +259,7 @@ This workflow carries out the steps in the [Standardising](#Standardising) workf
 | Parameter      | Type | Default | Description                                                                                                                                                     |
 | -------------- | ---- | ------- | --------------------------------------------------------------------------------------------------------------------------------------------------------------- |
 | ticket         | str  |         | Ticket ID e.g. 'AIP-55'                                                                                                                                         |
+| region         | enum |         | Region of the dataset :warning: The name has to be exactly one in the region enum in `standardising.                                                            |
```
amfage, line 262:
```suggestion
| region         | enum |         | Region of the dataset :warning: The name has to be exactly one in the region enum in `standardising`.                                                            |
```

workflows/basemaps/imagery-import-cogify.yml
```diff
@@ -35,6 +37,31 @@ spec:
...
+        value: 'new-zealand'
+        enum:
+          - 'antarctica'
+          - 'auckland'
+          - 'bay-of-plenty'
+          - 'canterbury'
+          - 'gisborne'
+          - 'global'
+          - 'hawkes-bay'
+          - 'manawatu-whanganui'
```
l0b0, line 51: Any particular reason to do the downcase, ASCII-only, hyphenated-rather-than-spaces transformation here? Wouldn't the users be better served with the official names, as normally written?

workflows/basemaps/imagery-import.yaml
```diff
@@ -21,6 +23,30 @@ spec:
       - name: ticket
         description: Ticket ID e.g. 'AIP-55'
         value: ''
+      - name: region
+        description: Region of the dataset
+        value: 'new-zealand'
+        enum:
```
l0b0, line 29: Would it be possible to de-dupe this parameter?

workflows/basemaps/imagery-import-cogify.yml
paulfouquet: This would add a lot of complexity to verify the values passed when submitting a workflow calling this template as I expect user to pass `-p region="hawkes-bay"` rather than `-p region="Hawkes Bay". I also don't think it is possible to transform these values with `labelsFrom`, as I would prefer having `linz.govt.nz/region=manawatu-whanganui` rather than `linz.govt.nz/region=Manawatū-Whanganui`. What are your thoughts?

workflows/basemaps/imagery-import.yaml
paulfouquet: From previous experience, I feel the default value HAS to be listed in the enum. That would make sense to control that `-p region="new-zealand"` is actually a valid region

l0b0: Wouldn't you just extract the parameter in its entirety, rather than just the enum?

workflows/basemaps/imagery-import-cogify.yml
l0b0: I'm probably not the best person to ask about this - I dislike transforming values to ASCII in general.

workflows/basemaps/imagery-import.yaml
paulfouquet: I am not sure to understand what you mean?

l0b0: Isn't it possible to create a `WorkflowTemplate` which defines only the `region` parameter, and import it in another `WorkflowTemplate`? For example:

```yaml
metadata:
  name: regions
  namespace: argo
spec:
  arguments:
    parameters:
      - name: region
        value: new-zealand
        enum:
          - antarctica
          - auckland
          - bay-of-plenty
          - canterbury
          - gisborne
          - global
          - hawkes-bay
          - manawatu-whanganui
          - marlborough
          - nelson
          - new-zealand
          - northland
          - otago
          - pacific-islands
          - southland
          - taranaki
          - tasman
          - waikato
          - wellington
          - west-coast
        description: Region of the dataset
```

l0b0: Actually, I couldn't figure out whether this is supported, so I've [asked around](https://stackoverflow.com/q/77567364/96588).

paulfouquet: [Templates](https://argoproj.github.io/argo-workflows/workflow-templates/) don't go in this level of details I think.

workflows/basemaps/imagery-import-cogify.yml
blacha: end-users hate writing utf8 in though, so anywhere its a "title", "description" or some form of text then 100% should be the official names, anywhere its a path or URL or domain name UTF8 just makes things break, especially really old geospatial software.

l0b0 (approved):

amfage (approved):

PR #343, opened 2023-12-14
feat: RDS usage metrics and alerts to Slack TDE-943

amfage:
#### Motivation

As a GDE developer, I want to retain CloudWatch logs and setup monitoring alarms so that I can monitor the health, usage and cost of the RDS instance and make sure that the Workflows Archive is reliable.

#### Modification

Add CloudWatch RDS metrics for available storage space and CPU utilisation. Add alerts to an SNS Topic which the AWS ChatBot will relay to the Argo dev alerts Slack channel.

The configuration for the Slack Channel is held in AWS Systems Manager Parameter Store. The AWS ChatBot requires approval for the Slack Channel (done via the AWS Console). This has been done.

#### Checklist

- [ ] Tests updated N/A
- [x] Docs updated (inline comments; not sure if more is needed).
- [X] Issue linked in Title


infra/eks/cluster.ts
```diff
@@ -97,6 +104,27 @@ export class LinzEksCluster extends Stack {
...
+    });
+    slackChannel.addNotificationTopic(rdsTopic);
+    const alarmStorage = this.argoDb.metricFreeStorageSpace().createAlarm(this, 'FreeStorageSpace', {
+      threshold: Size.gibibytes(2).toBytes(),
+      evaluationPeriods: 2,
+      comparisonOperator: cloudwatch.ComparisonOperator.LESS_THAN_THRESHOLD,
+    });
+    const alarmCpu = this.argoDb
+      .metricCPUUtilization()
+      .createAlarm(this, 'CPUUtilization', { threshold: 75, evaluationPeriods: 2 });
```
l0b0, line 124: Free storage space usually changes slowly over time, sometimes jumping up simply because of some process storing something temporary. So it makes sense for it to alert after only two evaluation periods. But CPU utilisation is usually much more random, and I worry that only two evaluation periods means we'll be getting a spurious alarm every few days or weeks. Something like 10 evaluation periods (assuming periods are 1 minute) would give us a much better indication that something is off. And considering our reaction time will probably average ~8 hours (if alarms randomly come outside of work hours), there shouldn't be anything important lost by increasing this. What do you think?

blacha: `.metricCPUUtilization()` has options for how the metric is set such as `period` which is the time that the stat is calculated which is defaulted to 5.

so this basically is average cpu over 10 minutes > 75% which seems pretty decent? 

It would be nice to put these values into the source code so we dont have to dig for what the default values are.

l0b0: Oh, OK. If it's average and five minutes per reading (I tried but couldn't find the reference for those defaults) then two periods sounds fine.

amfage: I have added a comment to give more information.
We can always adjust these settings in future if we need to.

blacha: Comments often get out of sync with the actual logic, I think its better to just add `period: Duration.minutes(5)` 

I don't think AWS would change these randomly, but :shrug: its not uncommon for people to change default values

l0b0: [`createAlarm`](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_cloudwatch.CreateAlarmOptions.html) doesn't take `period`. I think we'd have to set it on the code which generates the logs.

amfage: I think I would have to create the metric, set the period then set the alarm, which seemed like a bit to do to explicitly set something to the default.

blacha (approved):

l0b0 (approved):

PR #382, opened 2024-01-08
feat(basemaps): Move the create covering into spot instance to avoid exceeding memory

Wentao-Kuang:
#### Motivation

Create covering using too much memory for both processing and caching. We will remove the caching the future feature, and move the task into spot node to avoid exceed the memory.

#### Modification

Use spot instance for the create-covering task.

#### Checklist

- [ ] Tests updated
- [ ] Docs updated
- [ ] Issue linked in Title

### DIFF
--- a/workflows/basemaps/imagery-import-cogify.yml
+++ b/workflows/basemaps/imagery-import-cogify.yml
@@ -241,6 +241,8 @@ spec:
     - name: create-covering
       retryStrategy:
         limit: '2'
+      nodeSelector:
+        karpenter.sh/capacity-type: 'spot'
       inputs:
         parameters:
           - name: source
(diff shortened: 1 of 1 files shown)


blacha (approved):

l0b0 (commented):
I don't feel qualified to say whether this will work. How much memory was it getting before? And did you test this change with a big workload?

blacha:
> I don't feel qualified to say whether this will work. How much memory was it getting before? And did you test this change with a big workload?

It was running off the c6i.large base machines, which due to the argo logging bug are running low on memory, so this mostly just moves it off those base machines and onto spot instances.

Generally it only uses ~200MB of ram with 10,000 tiffs as input.

paulfouquet (approved):

PR #406, opened 2024-01-29
feat(basemaps): CRON workflow for the vector map etl. BM-950

Wentao-Kuang:
#### Motivation

Set up a cron workflow for the vector etl which is running 6am every Wednesday. So we can have up to date parcel and addressing data every week.

#### Modification
1. New cron workflow that reference the etl workflow.
2. Add a retry parameter for the existing workflow to allow cron workflow to retry automatically.

#### Checklist

- [ ] Tests updated
- [ ] Docs updated
- [ ] Issue linked in Title


workflows/basemaps/cron-vector-etl.yaml
```diff
@@ -0,0 +1,22 @@
+apiVersion: argoproj.io/v1alpha1
```
blacha, line 1:
```suggestion
# yaml-language-server: $schema=https://raw.githubusercontent.com/argoproj/argo-workflows/v3.4.13/api/jsonschema/schema.json
apiVersion: argoproj.io/v1alpha1
```

workflows/basemaps/cron-vector-etl.yaml
```diff
@@ -0,0 +1,22 @@
...
+metadata:
+  name: cron-vector-etl
+  labels:
+    linz.govt.nz/category: basemaps
+    linz.govt.nz/data-type: vector
+spec:
+  schedule: '0 06 * * 3'
+  timezone: 'NZ'
+  startingDeadlineSeconds: 3600 # Alow 1 hour delay if the workflow-controller clashes during the starting time.
+  concurrencyPolicy: 'Allow'
```
blacha, line 12: do we want runs to be concurrent?

Wentao-Kuang: You right, we should Forbid concurrent.

workflows/basemaps/cron-vector-etl.yaml
Wentao-Kuang: This is using for checking schema?

blacha: all the other files have this in, it mostly just adds autocomplete if you have the yaml extension installed.

workflows/basemaps/cron-vector-etl.yaml
l0b0: Neither of you have mentioned *why* concurrency should be allowed or not. Is it an established pattern not to allow it in this repo?

workflows/basemaps/cron-vector-etl.yaml
```diff
@@ -0,0 +1,23 @@
...
+kind: CronWorkflow
+metadata:
+  name: cron-vector-etl
+  labels:
+    linz.govt.nz/category: basemaps
+    linz.govt.nz/data-type: vector
+spec:
+  schedule: '0 06 * * 3'
+  timezone: 'NZ'
+  startingDeadlineSeconds: 3600 # Alow 1 hour delay if the workflow-controller clashes during the starting time.
```
l0b0, line 12:
```suggestion
  startingDeadlineSeconds: 3600 # Allow 1 hour delay if the workflow-controller clashes during the starting time.
```

workflows/basemaps/cron-vector-etl.yaml
blacha: In this case, I was asking if we want this cron job to run multiple instances of the cron job at the same time.

since it is only run once a week on a wednesday at 6am its very unlikely it would be ever trigger a concurrent run, but in the unlikely event it does it should not as its a singleton job.

workflows/basemaps/cron-vector-etl.yaml
```diff
@@ -0,0 +1,23 @@
...
+apiVersion: argoproj.io/v1alpha1
+kind: CronWorkflow
+metadata:
+  name: cron-vector-etl
+  labels:
+    linz.govt.nz/category: basemaps
+    linz.govt.nz/data-type: vector
+spec:
+  schedule: '0 06 * * 3'
+  timezone: 'NZ'
```
l0b0, line 11: "NZ" is not a valid time zone name. Based on the schema it looks like it should be this:

```suggestion
  timezone: 'Pacific/Auckland'
```

workflows/basemaps/cron-vector-etl.yaml
```diff
@@ -0,0 +1,23 @@
+# yaml-language-server: $schema=https://raw.githubusercontent.com/argoproj/argo-workflows/v3.4.13/api/jsonschema/schema.json
+apiVersion: argoproj.io/v1alpha1
+kind: CronWorkflow
+metadata:
+  name: cron-vector-etl
+  labels:
+    linz.govt.nz/category: basemaps
+    linz.govt.nz/data-type: vector
+spec:
+  schedule: '0 06 * * 3'
```
l0b0, line 10: Might be useful to add a comment like "6 AM every Wednesday"

workflows/basemaps/cron-vector-etl.yaml
```diff
@@ -0,0 +1,23 @@
...
+  timezone: 'NZ'
+  startingDeadlineSeconds: 3600 # Alow 1 hour delay if the workflow-controller clashes during the starting time.
+  concurrencyPolicy: 'Allow'
+  successfulJobsHistoryLimit: 3
+  failedJobsHistoryLimit: 1
```
l0b0, line 15: Wouldn't we want more information about failed jobs rather than successful jobs, so that we can compare subsequent failures?

workflows/basemaps/cron-vector-etl.yaml
Wentao-Kuang: Both NZ and Pacific/Auckland are IANA timezone standard, actually the NZ is links to Pacific/Auckland.

workflows/basemaps/vector-etl.yaml
```diff
@@ -30,6 +30,10 @@ spec:
           - 'true'
           - 'false'
 
+      - name: retry
+        description: Number of retry on failure vector-etl task
+        value: '1'
```
l0b0, line 35: How often do jobs fail in recoverable ways? That is, how often do we see a failed job which would succeed when re-run with the same parameters? If this is rare we should probably not enable retries.

workflows/basemaps/cron-vector-etl.yaml
l0b0: TIL, thanks!

workflows/basemaps/vector-etl.yaml
Wentao-Kuang: I would like to keep the auto reties. As the current ETL is not productionised yet, lack of testing. There might be changes failure sometimes. As this is cron job in the morning, auto retry will be great to have.

workflows/basemaps/cron-vector-etl.yaml
Wentao-Kuang: These are just the default setting, I haven't change. Both successful and failure job are quite useful. Successful have the output of s3 path. And failure got debug logs. As this only running once a week. What about we keep 3 of both successful and failure?

l0b0: > What about we keep 3 of both successful and failure?

That seems more useful. For example, in the case of a failure, we'd have two previous failures and three previous successes to compare to, which is very useful if anything has changed.

workflows/basemaps/cron-vector-etl.yaml
Wentao-Kuang: I think we can keep allow concurrency, as this won't bottle neck, and we might running multiple when debugging and testing.

workflows/basemaps/cron-vector-etl.yaml
Wentao-Kuang: I agree, especially when we doing testing. I have updated the failure history to 3. Thanks.

l0b0 (approved):

l0b0 (dismissed):

.github/workflows/main.yml
```diff
@@ -138,7 +138,15 @@ jobs:
           # Find all workflows that have kind "WorkflowTemplate"
           WORKFLOWS=$(grep '^kind: WorkflowTemplate$' -R workflows/ -H | cut -d ':' -f1)
 
-          # For each workflow attempt to deploy it using kubectl
+          # For each workflow template attempt to deploy it using kubectl
           for wf in $WORKFLOWS; do
               kubectl apply -f "$wf" --namespace argo
           done
+
+          # Find all cron workflows that have kind "CronWorkflow"
```
blacha, line 146: We need to deploy this to the test namespace too.

blacha (approved):

.github/workflows/main.yml
```diff
@@ -108,13 +108,22 @@ jobs:
...
-          # For each workflow attempt to deploy it using kubectl
+          # For each workflow template attempt to deploy it using kubectl
           for wf in $WORKFLOWS; do
               # Change namespace in files
               sed -i "/^\([[:space:]]*namespace: \).*/s//\1$GITHUB_SHA/" "$wf"
               kubectl apply -f "$wf" --namespace "$GITHUB_SHA"
           done
 
+          # Find all cron workflows that have kind "CronWorkflow"
+          CRON_WORKFLOWS=$(grep '^kind: CronWorkflow$' test/workflows/ | cut -d ':' -f1)
```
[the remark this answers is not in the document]
Wentao-Kuang: I am going to test this now, I just copy the same command from line 110. Probably they both not working?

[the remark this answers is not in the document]
l0b0: No, line 10 has those flags.

paulfouquet (approved):

paulfouquet:
Could initiate a documentation about the cron workflows in this or a separate PR

.github/workflows/main.yml
```diff
@@ -138,10 +147,15 @@ jobs:
...
-          # For each workflow attempt to deploy it using kubectl
+          # For each workflow template attempt to deploy it using kubectl
           for wf in $WORKFLOWS; do
               kubectl apply -f "$wf" --namespace argo
           done
 
-          # Deploy cron workflows
-          kubectl apply -f workflows/cron/ --namespace argo
+          # Find all cron workflows that have kind "CronWorkflow"
+          CRON_WORKFLOWS=$(grep '^kind: CronWorkflow$' -R workflows/ -H | cut -d ':' -f1)
```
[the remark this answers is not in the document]
Wentao-Kuang: I think we could update this in a separated PR, this PR is just for the cron workflow and copy the same commands for deployment. Another PR might be made to update all the commands.

l0b0 (approved):

PR #446, opened 2024-02-18
fix: create-manifest copy regex needs to match full stop

amfage:
#### Motivation

To prevent copy of the `.geojson` footprint files.

#### Modification

Escape the `.` in the `.json$` in the regex so we don't copy all `.geojson` files that get matched.

#### Checklist

N/A

- [ ] Tests updated
- [ ] Docs updated
- [ ] Issue linked in Title

### DIFF
--- a/workflows/raster/publish-odr.yaml
+++ b/workflows/raster/publish-odr.yaml
@@ -60,7 +60,7 @@ spec:
       - name: target_bucket_name
         value: 'nz-imagery'
       - name: include
-        value: '.tiff?$|.json$|^capture-area.geojson$'
+        value: '\.tiff?$|\.json$|^capture-area.geojson$'
       - name: copy_option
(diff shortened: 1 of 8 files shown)


paulfouquet (approved):

MDavidson17 (approved):

workflows/raster/publish-odr.yaml
```diff
@@ -60,7 +60,7 @@ spec:
       - name: target_bucket_name
         value: 'nz-imagery'
       - name: include
-        value: '.tiff?$|.json$|^capture-area.geojson$'
+        value: '\.tiff?$|\.json$|^capture-area.geojson$'
```
l0b0, line 63: Shouldn't the dot be escaped in `capture-area.geojson`?

PR #450, opened 2024-02-19
chore: update templates with Argo Tasks v3

amfage:
#### Motivation

Bring the templates into line with the workflows and workflowtemplates to be consistent. It should have no functional effect as the upstream workflows specify Argo Tasks `v3` which will override the templates, but there is always the possibility that this might not be specified upstream.

#### Modification

Change `version_argo_tasks` to `v3` in templates.

#### Checklist

- [ ] Tests updated
- [ ] Docs updated
- [ ] Issue linked in Title

### DIFF
--- a/templates/argo-tasks/copy.yml
+++ b/templates/argo-tasks/copy.yml
@@ -23,7 +23,7 @@ spec:
 
           - name: version_argo_tasks
             description: version of argo-tasks to use
-            default: 'v2'
+            default: 'v3'
 
           - name: copy_option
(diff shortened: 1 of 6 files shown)


paulfouquet (approved):

paulfouquet:
good finding

l0b0 (approved):
Is there some way we could automatically detect the mismatch here, for future changes? Or maybe it might be better not to specify the default, so that the users have to specify it explicitly?

PR #452, opened 2024-02-20
fix: regex for matching capture-area.geojson

amfage:
#### Motivation

The previous regex looked for a match for the start of the filename i.e. `^capture-area.geojson$`, however, when running on Argo Workflows, the path given when matching is always the full S3 object (path), so this did not match.

#### Modification

Change `^` to `/` to match the full filename. Also escape the full stop to look for a literal full stop character.

#### Checklist

N/A workflow bug

- [ ] Tests updated
- [ ] Docs updated
- [ ] Issue linked in Title

### DIFF
--- a/workflows/raster/copy.yaml
+++ b/workflows/raster/copy.yaml
@@ -3,7 +3,7 @@
 apiVersion: argoproj.io/v1alpha1
 kind: WorkflowTemplate
 metadata:
-  name: test-copy
+  name: copy
   namespace: argo
   labels:
     linz.govt.nz/category: raster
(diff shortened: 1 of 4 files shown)


workflows/raster/copy.yaml
```diff
@@ -3,7 +3,7 @@
 apiVersion: argoproj.io/v1alpha1
 kind: WorkflowTemplate
 metadata:
-  name: copy
+  name: test-copy
```
l0b0, line 6: This change looks unrelated. Is it?

amfage: Ah good catch thank you

l0b0 (approved):

MDavidson17 (approved):

PR #459, opened 2024-02-23
feat: enable publish to elevation TDE-1037

MDavidson17:
#### Motivation

Enable publish-odr to also push to linz/elevation 

#### Modification

**publish-odr:**
- Will always use a generated-target-path
- Will always publish to github
- Can publish to linz/imagery or linz/elevation

_nb: I will create a separate PR with a workflow that can be run in the event someone needs to publish to the odr and specifiy the target rather than automatically generating it (a backup workaround for any difficult datasets)_

**Copy**
- Will always require a target path (can use test-generate-path to build target)
- Will never publish to github
- Cannot publish to ODR

**test-generate-path**
New test workflow template for use to check the generated target path (superceding: https://github.com/linz/argo-tasks/blob/docs/tmp-generate-path/src/commands/path/TMP_DOC.md) 

#### Checklist

- [ ] Tests updated - unable to test publish-odr changes not in production?
- [x] Docs updated
- [x] Issue linked in Title

### DIFF
--- a/workflows/raster/README.md
+++ b/workflows/raster/README.md
@@ -244,7 +244,7 @@ Access permissions are controlled by the [Bucket Sharing Config](https://github.
 
 ## Workflow Description
 
-This workflow replicates `copy` however it allows publishing to `s3://nz-imagery` (the registry of open data).
+This workflow replicates `copy` however it allows publishing to `s3://nz-imagery` and `s3://nz-elevation` (the two registries of open data).
 **This workflow should not be run using the Argo UI, instead follow the instruction [here](https://github.com/linz/imagery/tree/master/publish-odr-parameters/README.md)**
 
 ```mermaid
--- a/workflows/raster/README.md
+++ b/workflows/raster/README.md
@@ -191,14 +191,10 @@ Copy files from one S3 location to another. This workflow is intended to be used
 
 ```mermaid
 graph TD;
-  create-manifest-->copy-.->push-to-github;
+  create-manifest-->copy;
 ```
 
-\* `push-to-github` is an optional task run only for `s3://linz-imagery/`
-
 This is a workflow that uses the [argo-tasks](https://github.com/linz/argo-tasks#create-manifest) container `create-manifest` (list of source and target file paths) and `copy` (the actual file copy) commands.
-
-If the target destination is `linz-imagery` there is an extra step carried out that pushes the `collection.json` to the `linz/imagery` GitHub repository, which is where the STAC Catalog and Collection metadata files for the `s3://linz-imagery` bucket are managed. This step uses the argo-tasks container `stac-github-import` command.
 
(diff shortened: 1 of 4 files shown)


l0b0 (dismissed):

workflows/raster/README.md
```diff
@@ -244,7 +240,7 @@ Access permissions are controlled by the [Bucket Sharing Config](https://github.
 
 ## Workflow Description
 
-This workflow replicates `copy` however it allows publishing to `s3://nz-imagery` (the registry of open data).
+This workflow replicates `copy` however it allows publishing to `s3://nz-imagery` and `s3://nz-elevation` (the two registries of open data).
```
amfage, line 243:
```suggestion
This workflow replicates `copy` however it allows publishing to `s3://nz-imagery` and `s3://nz-elevation` (the two registry of open data public S3 buckets).
```

workflows/test/generate-path.yaml
```diff
@@ -0,0 +1,45 @@
...
+      - name: version_argo_tasks
+        value: 'v3'
+      - name: source
+        value: 's3://linz-imagery-staging/test/sample/'
+      - name: target_bucket_name
+        value: 'nz-imagery'
+  templateDefaults:
+    container:
+      imagePullPolicy: Always
+      image: ''
```
l0b0, line 25: Won't this break if we don't override it?

MDavidson17: I'm not sure what you mean or of the answer, I copied these defaults from the other test templates. 
I have run it in argo and it worked: https://argo.linzaccess.com/workflows/argo/test-generate-target-52mg6?tab=workflow

amfage (approved):

workflows/test/generate-path.yaml
l0b0: My assumptions:

- An empty string is not a valid image reference.
- A workflow run with an invalid image reference will always fail.
- The difference between running with an invalid image reference and *no* image reference is not important.

Some scenarios we need to consider:

1. The empty value overrides some other default higher up in the stack of templates. By setting it to an empty string we *force* users of this template to specify a different value, to avoid running with an invalid image reference. This seems like a bad hack, and unlikely.
2. An empty image value has some sort of special significance in Argo. This seems really unlikely, as it would be horrible design.
3. We can's specify an `imagePullPolicy` default without also specifying an `image` default. This would be a good reason to leave the line in, but seems unlikely.
4. Specifying an image default is unnecessary, because it's always overridden.

The last scenario seems the most likely, so I expect removing this line is better than leaving it in.

l0b0 (commented):
I responded to your comment.

l0b0 (approved):

workflows/test/generate-path.yaml
MDavidson17: We've decided to pull this out into another ticket, it would be good to look into this but there are multiple instances of this code block so we have to do some testing.

PR #484, opened 2024-03-14
feat: save basemaps config url as artifact in s3 TDE-1098

paulfouquet:
#### Motivation

Saving the Basemaps config URL in the s3 target path (flat directory) will allow another workflow to grab its value.

#### Modification

Save an artifact config-url containing the basemaps config url in s3

#### Checklist

- [ ] Tests updated N/A
- [ ] Docs updated N/A
- [x] Issue linked in Title

### DIFF
--- a/workflows/raster/standardising.yaml
+++ b/workflows/raster/standardising.yaml
@@ -3,7 +3,7 @@
 apiVersion: argoproj.io/v1alpha1
 kind: WorkflowTemplate
 metadata:
-  name: imagery-standardising
+  name: test-config-url-artifact
   namespace: argo
   labels:
     linz.govt.nz/category: raster
@@ -381,6 +381,10 @@ spec:
               parameters:
                 - name: location
                   value: '{{tasks.get-location.outputs.parameters.location}}'
+                - name: bucket
+                  value: '{{tasks.get-location.outputs.parameters.bucket}}'
+                - name: key
+                  value: '{{tasks.get-location.outputs.parameters.key}}'
             template: create-config
             depends: 'create-overview'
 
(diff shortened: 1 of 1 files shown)


workflows/raster/standardising.yaml
```diff
@@ -594,6 +608,14 @@ spec:
             description: Location of the config file
             valueFrom:
               path: '/tmp/cogify/config-path'
+        artifacts:
+          - name: url
```
paulfouquet, line 612: It is still useful to keep the output parameter `url` as it allows to grab the url from a workflow viewed from Argo Workflow UI without looking at the logs or downloading the `url` artifact.

amfage (approved):

workflows/raster/standardising.yaml
```diff
@@ -536,11 +540,19 @@ spec:
           const loc = JSON.parse(process.env['ARGO_TEMPLATE']).archiveLocation.s3;
           const key = loc.key.replace('{{pod.name}}','');
           fs.writeFileSync('/tmp/location', `s3://${loc.bucket}/${key}`);
+          fs.writeFileSync('/tmp/bucket', `${loc.bucket}`);
+          fs.writeFileSync('/tmp/key', `${key}`);
```
l0b0, line 544: Looks we shouldn't need `location` anymore, since it can be constructed from `bucket` and `key`…

paulfouquet: `location` is used in few places and needs to be construct. I did not want to modify the existing to make it more complicated.

l0b0 (approved):

PR #485, opened 2024-03-18
feat: update publish-odr to remove copy and add new PR fields TDE-1100 TDE-1102

amfage:
#### Motivation

Provide more info for imagery and elevation PR reviews and copy the item TIFFs and STAC files only after the PR is approved and merged.

#### Modification

The ticket and copy-option need to be part of the Pull Request.
The `create-manifest` and `copy` tasks are removed from the `publish-odr` workflowtemplate.
The `publish-to-github` workflowtemplate is updated for the new fields in Argo Tasks that are to be passed through to the Pull Request.

#### Checklist

- [ ] Tests updated
- [x] Docs updated
- [x] Issue linked in Title

### DIFF
--- a/templates/argo-tasks/README.md
+++ b/templates/argo-tasks/README.md
@@ -161,12 +161,16 @@ See https://github.com/linz/argo-tasks#stac-github-import
       - name: source
         value: '{{inputs.parameters.source}}'
       - name: target
-        value: '{{workflow.parameters.target}}'
+        value: '{{tasks.generate-path.outputs.parameters.target}}'
       - name: version_argo_tasks
         value: '{{workflow.parameters.version_argo_tasks}}'
       - name: repository
-        value: 'elevation'
-  depends: 'copy-with-github'
+        value: "{{=sprig.trimPrefix('nz-', workflow.parameters.target_bucket_name)}}"
+      - name: ticket
+        value: '{{=sprig.trim(workflow.parameters.ticket)}}'
+      - name: copy_option
+        value: '{{workflow.parameters.copy_option}}'
+  depends: 'generate-path'
 ```
 
 ## argo-tasks/generate-path
--- a/templates/argo-tasks/push-to-github.yml
+++ b/templates/argo-tasks/push-to-github.yml
@@ -36,7 +36,7 @@ spec:
 
           - name: ticket
             description: Ticket ID e.g. 'AIP-55'
-            value: ''
+            default: ''
 
           - name: copy_option
             description: --no-clobber Skip overwriting existing files. --force Overwrite all files. --force-no-clobber Overwrite only changed files, skip unchanged files.
--- a/templates/argo-tasks/push-to-github.yml
+++ b/templates/argo-tasks/push-to-github.yml
@@ -29,7 +29,7 @@ spec:
 
           - name: repository
             description: Repository Name (only used when identified that push to GitHub is required)
-            default: 'elevation'
+            default: 'imagery'
             enum:
               - 'elevation'
               - 'imagery'
@@ -58,13 +58,13 @@ spec:
             'stac',
             'github-import',
             '--source',
-            '{{=sprig.trim(inputs.parameters.source)}}',
+            '{{inputs.parameters.source}}',
             '--target',
-            '{{=sprig.trim(inputs.parameters.target)}}',
+            '{{inputs.parameters.target}}',
             '--repo-name',
             'linz/{{=sprig.regexFind(inputs.parameters.repository, inputs.parameters.target)}}',
             '--ticket',
-            '{{=sprig.trim(inputs.parameters.target)}}',
+            '{{inputs.parameters.ticket}}',
             '--copy-option',
-            '{{=sprig.trim(inputs.parameters.copy_option)}}',
+            '{{inputs.parameters.copy_option}}',
           ]
--- a/templates/argo-tasks/push-to-github.yml
+++ b/templates/argo-tasks/push-to-github.yml
@@ -34,6 +34,14 @@ spec:
               - 'elevation'
               - 'imagery'
 
+          - name: ticket
+            description: Ticket ID e.g. 'AIP-55'
+            value: ''
+
+          - name: copy_option
+            description: --no-clobber Skip overwriting existing files. --force Overwrite all files. --force-no-clobber Overwrite only changed files, skip unchanged files.
+            default: '--no-clobber'
+
       container:
         image: '019359803926.dkr.ecr.ap-southeast-2.amazonaws.com/argo-tasks:{{=sprig.trim(inputs.parameters.version_argo_tasks)}}'
         env:
@@ -55,4 +63,8 @@ spec:
             '{{=sprig.trim(inputs.parameters.target)}}',
             '--repo-name',
             'linz/{{=sprig.regexFind(inputs.parameters.repository, inputs.parameters.target)}}',
+            '--ticket',
+            '{{=sprig.trim(inputs.parameters.target)}}',
+            '--copy-option',
+            '{{=sprig.trim(inputs.parameters.copy_option)}}',
           ]
--- a/templates/argo-tasks/push-to-github.yml
+++ b/templates/argo-tasks/push-to-github.yml
@@ -29,14 +29,12 @@ spec:
 
           - name: repository
             description: Repository Name
-            default: ''
             enum:
               - 'elevation'
               - 'imagery'
 
           - name: ticket
             description: Ticket ID e.g. 'AIP-55'
-            default: ''
 
           - name: copy_option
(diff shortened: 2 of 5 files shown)


templates/argo-tasks/README.md
```diff
@@ -161,12 +161,16 @@ See https://github.com/linz/argo-tasks#stac-github-import
...
         value: '{{inputs.parameters.source}}'
       - name: target
-        value: '{{workflow.parameters.target}}'
+        value: '{{tasks.generate-path.outputs.parameters.target}}'
       - name: version_argo_tasks
         value: '{{workflow.parameters.version_argo_tasks}}'
       - name: repository
-        value: 'elevation'
-  depends: 'copy-with-github'
+        value: "{{=sprig.trimPrefix('nz-', workflow.parameters.target_bucket_name)}}"
```
l0b0, line 168: Is there a way to use a dictionary lookup instead? That would make the code a bit more future-proof, by making it easy to add arbitrary bucket name ↔ repository name pairs.

templates/argo-tasks/push-to-github.yml
```diff
@@ -28,12 +28,20 @@ spec:
             default: 'v3'
 
           - name: repository
-            description: Repository Name (only used when identified that push to GitHub is required)
-            default: 'elevation'
+            description: Repository Name
+            default: 'imagery'
```
l0b0, line 32: Do we need to specify a default? If there's approximately equal numbers of elevation and imagery submissions we could just not specify one.

templates/argo-tasks/push-to-github.yml
```diff
@@ -28,12 +28,20 @@ spec:
...
             enum:
               - 'elevation'
               - 'imagery'
 
+          - name: ticket
+            description: Ticket ID e.g. 'AIP-55'
+            value: ''
```
l0b0, line 39: Should this have a value? Is it not required?

workflows/raster/README.md
```diff
@@ -240,12 +240,12 @@ Access permissions are controlled by the [Bucket Sharing Config](https://github.
 
 ## Workflow Description
 
-This workflow replicates `copy` however it allows publishing to `s3://nz-imagery` and `s3://nz-elevation` (the two registry of open data public S3 buckets).
-**This workflow should not be run using the Argo UI, instead follow the instruction [here](https://github.com/linz/imagery/tree/master/publish-odr-parameters/README.md)**
+This workflow creates a GitHub Pull Request to be reviewed for publishing to `s3://nz-imagery` and `s3://nz-elevation` (the two registry of open data public S3 buckets). When the Pull Request is approved and merged, the files will be copied.
```
l0b0, line 243: Is "registry of open data" here referring to https://registry.opendata.aws/? If so it should probably be capitalised. "Pull Request", on the other hand, should not be capitalised.

templates/argo-tasks/README.md
```diff
@@ -161,12 +161,16 @@ See https://github.com/linz/argo-tasks#stac-github-import
       - name: source
         value: '{{inputs.parameters.source}}'
       - name: target
-        value: '{{workflow.parameters.target}}'
+        value: '{{tasks.generate-path.outputs.parameters.target}}'
```
paulfouquet, line 164: I wonder if these values in this example of how to use the template should just be placeholders, rather than having to update them when the "real" example changes.

templates/argo-tasks/push-to-github.yml
paulfouquet, line 32: Why did we change from `elevation` to `imagery` for the default value? Should we use an ABC order instead? Also should we use enums?

templates/argo-tasks/push-to-github.yml
```diff
@@ -28,12 +28,20 @@ spec:
             default: 'v3'
 
           - name: repository
-            description: Repository Name (only used when identified that push to GitHub is required)
-            default: 'elevation'
+            description: Repository Name
+            default: 'imagery'
             enum:
```
paulfouquet, line 33: Is the enum here to inform user what we can use? because I don't think there is any user control on these

templates/argo-tasks/push-to-github.yml
```diff
@@ -28,12 +28,20 @@ spec:
...
               - 'elevation'
               - 'imagery'
 
+          - name: ticket
+            description: Ticket ID e.g. 'AIP-55'
+            value: ''
+
+          - name: copy_option
+            description: --no-clobber Skip overwriting existing files. --force Overwrite all files. --force-no-clobber Overwrite only changed files, skip unchanged files.
+            default: '--no-clobber'
```
paulfouquet, line 43: Do we use enums like the repository name, or not use enum for both?

templates/argo-tasks/push-to-github.yml
amfage: There is no default value for this field.

templates/argo-tasks/README.md
MDavidson17: I have noted this code as a [bug](https://toitutewhenua.atlassian.net/browse/TDE-1053?atlOrigin=eyJpIjoiNzdhNGYzNWRkNWRmNGRlM2FlOGNhNWRhMGYzMmQ0NTgiLCJwIjoiaiJ9).

If you fix it here feel free to close this bug, or if it is complex, we could fix it in a separate PR.

templates/argo-tasks/push-to-github.yml
l0b0: The [docs](https://argo-workflows.readthedocs.io/en/latest/fields/#parameter) say

> Value is the literal value to use for the parameter. If specified in the context of an input parameter, the value takes precedence over any passed values

Which means this will always have an empty value. I don't think that's what we want.

templates/argo-tasks/push-to-github.yml
amfage: Maybe there to make it more obvious? As you say, it will not have any other effect.

templates/argo-tasks/push-to-github.yml
amfage: Ah yes it should be `default` instead of `value`.

l0b0: Why use a default of empty string though? Don't we want this to be a mandatory field?

amfage: I've removed the defaults; they are always overridden.

workflows/raster/publish-odr.yaml
```diff
@@ -58,21 +58,16 @@ spec:
       - name: source
         value: 's3://linz-imagery-staging/test/sample/'
       - name: target_bucket_name
-        value: 'nz-imagery'
-      - name: include
-        value: '\.tiff?$|\.json$|/capture-area\.geojson$'
+        value: ''
```
l0b0, line 61: Isn't this also broken, in that the value will always be an empty string?

amfage: This is an argument parameter rather than an input parameter so will be overridden by the workflow parameters.

templates/argo-tasks/README.md
paulfouquet: As discussed, this is not so important for this ticket

templates/argo-tasks/push-to-github.yml
```diff
@@ -28,12 +28,17 @@ spec:
...
-            default: 'elevation'
+            description: Repository Name
             enum:
               - 'elevation'
               - 'imagery'
 
+          - name: ticket
+            description: Ticket ID e.g. 'AIP-55'
+
+          - name: copy_option
```
paulfouquet, line 39: To be consistent with `repository` we could use `enum` here too

templates/argo-tasks/README.md
amfage: Ticket for reviewing topo-workflows documentation: https://toitutewhenua.atlassian.net/browse/TDE-1104

l0b0 (approved):

templates/argo-tasks/push-to-github.yml
amfage: I've removed all the enums from this file now, they have no functional purpose as they are never submitted through the UI.

templates/argo-tasks/push-to-github.yml
amfage: I've removed all the enums from this file now, they have no functional purpose as they are never submitted through the UI.

templates/argo-tasks/push-to-github.yml
amfage: I've removed all the enums from this file now, they have no functional purpose as they are never submitted through the UI.

paulfouquet (approved):

l0b0 (approved):

PR #491, opened 2024-03-20
feat: Disable IPv6 for Fluent Bit TDE-1093

l0b0:
#### Motivation

_What does this change aim to achieve?_

#### Modification

_Why is this change being made? What implications or other considerations are there?_

#### Checklist

- [ ] Tests updated
- [ ] Docs updated
- [x] Issue linked in Title

### DIFF
--- a/infra/charts/fluentbit.ts
+++ b/infra/charts/fluentbit.ts
@@ -54,6 +54,7 @@ export class FluentBit extends Chart {
      *
      */
     const extraService = `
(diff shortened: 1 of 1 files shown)


blacha:
Do you think we still need this?

From the docs:
> Prioritize IPv4 DNS results when trying to establish a connection.

Anything inside the cluster will only return a IPV6 so it cant use IPV4, anything outside the cluster would only return a IPV4 so it doesn't need to choose there either?

PR #499, opened 2024-03-28
feat: force fqdn dns request for cloudwatch resolution by fluentbit TDE-1093

paulfouquet:
#### Motivation

Inside kubernetes, DNS resolution uses ndots: 5 this means any dns requests that has less than 5 dots will go through all the search domains before looking up the actual domain. AWS Cloudwatch `logs.ap-southeast-2.amazonaws.com` has 3 dots.

#### Modification

Hardcode the Cloudwatch endpoint  to `logs.ap-southeast-2.amazonaws.com.` so DNS requests to Cloudwatch force the DNS lookup to be fully qualified.
This has been tested on a non prod cluster and we've seen a reduction of the DNS queries to the `coredns`

#### Checklist

- [ ] Tests updated
- [x] Docs updated
- [x] Issue linked in Title

### DIFF
--- a/infra/charts/fluentbit.ts
+++ b/infra/charts/fluentbit.ts
@@ -74,6 +74,8 @@ HC_Period 5
         cloudWatchLogs: {
           enabled: true,
           region: 'ap-southeast-2',
+          /** Specify Cloudwatch endpoint to add a trailing `.` to force FQDN DNS request */
+          endpoint: 'logs.ap-southeast-2.amazonaws.com.',
           autoCreateGroup: true,
           logRetentionDays: 30,
           logGroupName: `/aws/eks/${props.clusterName}/logs`,


infra/charts/fluentbit.ts
```diff
@@ -74,6 +74,8 @@ HC_Period 5
         cloudWatchLogs: {
           enabled: true,
           region: 'ap-southeast-2',
```
paulfouquet, line 76: I am wondering if this is only used to build the endpoint, and as we force the endpoint we probably don't need it anymore. But let's leave it just in case?

l0b0 (approved):

infra/charts/fluentbit.ts
```diff
@@ -74,6 +74,8 @@ HC_Period 5
...
+          /** Specify Cloudwatch endpoint to add a trailing `.` to force FQDN DNS request */
+          endpoint: 'logs.ap-southeast-2.amazonaws.com.',
```
l0b0, line 78: Nit: We should probably pull out a variable for the region, and use that in here, so that it's consistent.

paulfouquet: Could store in a variable the region in a separate pull request if there are other spots we use it.

amfage (approved):

l0b0 (approved):

PR #511, opened 2024-04-04
feat: add node local dns cache

blacha:
#### Motivation

We are seeing a large amount of DNS failures with timeouts and `EAI_AGAIN` responses, we believe it might be due to the DNS servers getting overwhelmed when the cluster is scalling rapidly up.

By adding a DNS cache to every node there is a huge reduction of DNS requests made over the wire to the primary coredns servers.

#### Modification

Adds a DNS cache to every node inside the cluster.

### DIFF
--- a/docs/dns.configuration.md
+++ b/docs/dns.configuration.md
@@ -69,7 +69,7 @@ Depending on the container you may have access to scripting languages.
 
 #### NodeJS
 
-create a new file  `index.mjs`
+create a new file `index.mjs`
 
 ```javascript
 fetch('https://google.com').then((c) => console.log(c));
@@ -80,7 +80,7 @@ console.log(await dns.resolve('google.com', 'A'));
 console.log(await dns.resolve('google.com', 'AAAA'));
 ```
 
-Run the file 
+Run the file
 
 ```bash
 node --version
--- a/docs/dns.configuration.md
+++ b/docs/dns.configuration.md
@@ -15,7 +15,7 @@ k exec -n :namespace -it :podName -- /bin/bash
 Install basic dns utils `dig` `ping` `wget` and `curl`
 
 ```bash
-apt install dnsutils iptools-ping wget curl
+apt update && apt install dnsutils wget curl -y
 ```
 
 ### Name resolution
@@ -69,18 +69,24 @@ Depending on the container you may have access to scripting languages.
 
 #### NodeJS
 
-file: index.mjs
+create a new file  `index.mjs`
 
 ```javascript
 fetch('https://google.com').then((c) => console.log(c));
 
 import * as dns from 'dns/promises';
 
-await dns.resolve('google.com', 'A');
-await dns.resolve('google.com', 'AAAA');
+console.log(await dns.resolve('google.com', 'A'));
+console.log(await dns.resolve('google.com', 'AAAA'));
 ```
+
+Run the file 
 
 ```bash
 node --version
 node index.mjs
 ```
+
+## Node Local DNS
+
+A local DNS cache is running on every node, [node-local-dns](./infrastructure/components/node.local.dns.md) if any DNS issues occur it is recommended to turn the DNS cache off as a first step for debugging
--- a/docs/dns.configuration.md
+++ b/docs/dns.configuration.md
@@ -12,10 +12,16 @@ Start a shell on the container
 k exec -n :namespace -it :podName -- /bin/bash
 ```
 
-Install basic dns utils `dig` `ping` `wget` and `curl`
+Install basic networking utils `dig`, `ping`, `ping6`, `wget`, `nslookup`, and `curl`
 
 ```bash
-apt update && apt install dnsutils wget curl -y
+apt update && apt install -y dnsutils iputils-ping wget curl
+```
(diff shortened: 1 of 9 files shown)


docs/infrastructure/components/node.local.dns.md
```diff
@@ -0,0 +1,42 @@
...
+
+By default the DNS cache will log any external DNS requests it is resolving (anything that is not ending with `.cluster.local`) since there can be a large number of dns cache pods the following command will tail the logs from
+
+```
+kubectl logs -n kube-system --all-containers=true -f daemonset/node-local-dns --since=1m --timestamps=true --prefix=true
+```
+
+### Structured logs
+
+coredns does not by provide a simple way of constructing a structured log from the DNS request, it does provide a template system which can be used to craft a JSON log line, if the log line is in structured format like JSON it can be more easily processed into something like elasticsearch for additional debugging
```
paulfouquet, line 23:
```suggestion
`coredns` does not provide a simple way of constructing a structured log from the DNS request, it does provide a template system which can be used to craft a JSON log line, if the log line is in structured format like JSON it can be more easily processed into something like elasticsearch for additional debugging.
```

docs/dns.configuration.md
```diff
@@ -15,7 +15,7 @@ k exec -n :namespace -it :podName -- /bin/bash
 Install basic dns utils `dig` `ping` `wget` and `curl`
 
 ```bash
-apt install dnsutils iptools-ping wget curl
+apt update && apt install dnsutils wget curl -y
```
l0b0, line 18: After running this on the latest official Ubuntu container there's no `ping` command.

infra/cdk8s.ts
```diff
@@ -7,32 +7,45 @@ import { EventExporter } from './charts/event.exporter.js';
...
+import { CfnOutputKeys, ClusterName, ScratchBucketName, UseNodeLocalDns, validateKeys } from './constants.js';
+import { describeCluster, getCfnOutputs } from './util/cloud.formation.js';
 import { fetchSsmParameters } from './util/ssm.js';
 
 const app = new App();
 
 async function main(): Promise<void> {
   // Get cloudformation outputs
-  const cfnOutputs = await getCfnOutputs(ClusterName);
+  const [cfnOutputs, ssmConfig, clusterConfig] = await Promise.all([
```
l0b0, line 19: Why wrap these in `await Promise.all()`?

infra/charts/kube-system.node.local.dns.ts
```diff
@@ -0,0 +1,200 @@
...
+            '-localip',
+            [bindAddress, upstreamDns].join(','),
+            '-conf',
+            '/etc/Corefile',
+            '-upstreamsvc',
+            dnsUpstream.name,
+          ],
+          ports: [
+            { name: 'dns', protocol: kplus.Protocol.UDP, number: 53 },
+            // { name: 'dns-tcp', protocol: kplus.Protocol.TCP, number: 53 }, // TODO this is broken, see JSONPatch
```
l0b0, line 94: Do you have a reference to the bug? Otherwise it's really hard to know when to get rid of the workaround.

infra/cdk8s.ts
blacha: It runs all the queries concurrently rather than sequentially, cuts time to run these step down by approx 1/3rd 120ms+150ms+100ms -> ~150ms

infra/charts/kube-system.node.local.dns.ts
```diff
@@ -0,0 +1,200 @@
...
+      }
+      reload
+      loop
+      bind ${ctx.bindAddress} ${ctx.upstreamAddress}
+      forward . __PILLAR__CLUSTER__DNS__ {
+              force_tcp
+      }
+      prometheus :9253
+      health [${ctx.bindAddress}]:8080
+      }
```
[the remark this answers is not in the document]
blacha: this was taken from https://github.com/kubernetes/kubernetes/blob/master/cluster/addons/dns/nodelocaldns/nodelocaldns.yaml#L56 and then some values replaced with the known values, it could be deindented I think its only the trailing `}` needing to be de-indented.

infra/charts/kube-system.node.local.dns.ts
blacha: I couldn't find a bug for this issue (or any of the issues which are currently JSONPath'd) their docs only say most of the apis are implemented.

I tried upgrading to the latest cdk8s but still couldn't use TCP & UDP in the same port, It could also just mean I'm doing it not in the cdk8s way :shrug: ? .

docs/dns.configuration.md
blacha: found it at `iputils-ping`, I dont think `iptools-ping` is a actual package weird..

infra/charts/kube-system.node.local.dns.ts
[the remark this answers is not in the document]
blacha: and de-indented with some extra docs about where this file came from.

paulfouquet (approved):

l0b0 (approved):

PR #529, opened 2024-04-15
ci: move workflows linting to the main job TDE-1143

paulfouquet:
#### Motivation

We are not linting our workflows at any time. The linting should be done on any push before merging to master.

#### Modification

Move the linting task from the deploy-prod job to the main job

#### Checklist

- [ ] Tests updated
- [ ] Docs updated
- [x] Issue linked in Title

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,6 +1,9 @@
 on: [push]
 env:
   CLUSTER_NAME: Workflows
+permissions:
+  id-token: write
+  contents: read
 
 jobs:
   main:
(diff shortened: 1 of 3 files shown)


blacha:
I wonder if there is a better way of handling this? is there anyway to get argo to lint against local yaml files?

this would make the github action for every push login and deploy stuff into our production argo cluster, with only a bit of bash and sed preventing it from accidentally deploying into the actual production workloads.

paulfouquet:
https://github.com/linz/topo-workflows/pull/540 replaced this PR

PR #530, opened 2024-04-15
refactor: templating stac-validate workflow TDE-1136

paulfouquet:
#### Motivation

`stac-validate` is used in many places, it should be easier to maintain and reuse if it was a "template".

#### Modification

Create a `stac-validate` template so it can be re-use by workflows

#### Checklist

- [ ] Tests updated
- [x] Docs updated
- [x] Issue linked in Title

### DIFF
--- a/templates/argo-tasks/stac-validate.yml
+++ b/templates/argo-tasks/stac-validate.yml
@@ -47,11 +47,9 @@ spec:
           - name: AWS_ROLE_CONFIG_PATH
             value: s3://linz-bucket-config/config.json
         args:
-          [
-            'stac',
-            'validate',
-            '--concurrency={{inputs.parameters.concurrency}}',
-            '--recursive={{inputs.parameters.recursive}}',
-            '--checksum={{inputs.parameters.checksum}}',
-            '{{inputs.parameters.uri}}',
-          ]
+            - 'stac'
+            - 'validate'
+            - '--concurrency={{inputs.parameters.concurrency}}'
+            - '--recursive={{inputs.parameters.recursive}}'
+            - '--checksum={{inputs.parameters.checksum}}'
+            - '{{inputs.parameters.uri}}'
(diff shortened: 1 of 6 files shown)


templates/argo-tasks/stac-validate.yml
```diff
@@ -0,0 +1,57 @@
...
+        resources:
+          requests:
+            cpu: 15000m
+            memory: 7.8Gi
+        command: [node, /app/index.js]
+        env:
+          - name: AWS_ROLE_CONFIG_PATH
+            value: s3://linz-bucket-config/config.json
+        args:
+          [
```
blacha, line 50: NIT: do we have a preference on `[ .. ]` vs `- - -` style for arrays?

templates/argo-tasks/stac-validate.yml
```diff
@@ -0,0 +1,57 @@
...
+
+          - name: version
+            description: container version to use
+            default: 'v3'
+
+      container:
+        image: '019359803926.dkr.ecr.ap-southeast-2.amazonaws.com/argo-tasks:{{=sprig.trim(inputs.parameters.version)}}'
+        resources:
+          requests:
+            cpu: 15000m
```
blacha, line 43: Is there anyway to scale this differently if `--checksum` is enabled, we do not need 15CPUs to validate a few JSON files?

l0b0 (dismissed):

templates/argo-tasks/stac-validate.yml
l0b0: We should probably do that in a separate ticket, rather than changing two unrelated things in this one.

paulfouquet: We should probably do it in this change since we are making it a generic template.

l0b0 (dismissed):

templates/argo-tasks/stac-validate.yml
paulfouquet: @blacha It does not seem possible to [configure the pod resources using parameters](https://github.com/argoproj/argo-workflows/issues/7839). I would not want [to use `podSpecPatch`](https://github.com/argoproj/argo-workflows/issues/3058#issuecomment-631206170) in this pull request.

blacha (approved):

MDavidson17 (approved):

MDavidson17 (approved):

l0b0 (approved):

PR #540, opened 2024-04-17
ci: use --offline mode to lint workflows TDE-1149 TDE-1143

paulfouquet:
#### Motivation

The current process of linting the workflows is complex as we need to publish them on the kubernetes cluster. Using an offline mode makes things simpler and quicker, removing the connection to the cluster.

#### Modification

Use the `--offline` mode to run `argo lint`, rather than linting from the resources deployed on the cluster.

#### Checklist

- [ ] Tests updated
- [ ] Docs updated
- [X] Issue linked in Title

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,9 +1,4 @@
 on: [push]
-env:
-  CLUSTER_NAME: Workflows
-permissions:
-  id-token: write
-  contents: read
 
 jobs:
   main:
@@ -26,7 +21,6 @@ jobs:
           ./argo-linux-amd64 version
 
       - name: Lint workflows
-        if: github.ref != 'refs/heads/master'
         run: |
           ./argo-linux-amd64 lint --offline templates/ workflows/
 
@@ -38,6 +32,13 @@ jobs:
 
     environment:
       name: prod
+
+    permissions:
+      id-token: write
+      contents: read
+
+    env:
+      CLUSTER_NAME: Workflows
 
     steps:
       - uses: linz/action-typescript@9bf69b0f313b3525d3ba3116f26b1aff7eb7a6c0 # v3.1.0
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,4 +1,9 @@
 on: [push]
+env:
+  CLUSTER_NAME: Workflows
+permissions:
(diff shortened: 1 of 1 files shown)


l0b0 (dismissed):
:tada:

MDavidson17 (dismissed):

.github/workflows/main.yml
```diff
@@ -1,4 +1,9 @@
 on: [push]
+env:
+  CLUSTER_NAME: Workflows
+permissions:
```
blacha, line 4: Why are we increasing the permissions for all steps?

.github/workflows/main.yml
```diff
@@ -13,6 +18,18 @@ jobs:
       - name: Run actionlint to check workflow files
         run: docker run --volume="${PWD}:/repo" --workdir=/repo actionlint -color
 
+      - name: Install Argo
```
blacha, line 21: if we arent going to use argo on `refs/heads/master` there isnt much point downloading it?

How fast/slow is lint could we just lint everytime?

.github/workflows/main.yml
paulfouquet: Oops that's an artifact from [the previous change](https://github.com/linz/topo-workflows/pull/529) that I copied past. I'll fix that.

.github/workflows/main.yml
paulfouquet: > if we arent going to use argo on `refs/heads/master` there isnt much point downloading it?

I am not sure what you mean about this. We need to lint them on push when a pull request is made. 

> How fast/slow is lint could we just lint everytime?

I don't have the numbers, but it is fast.

blacha: > I am not sure what you mean about this. We need to lint them on push when a pull request is made.

The download always runs, but the next step is conditional https://github.com/linz/topo-workflows/pull/540/files/7b883e91dd813dfccbba765b556dc534a30a2cb5#diff-7829468e86c1cc5d5133195b5cb48e1ff6c75e3e9203777f6b2e379d9e4882b3R29

so on merge to master it will download, but then skip the next step

I think I would prefer we just remove L29. so the lint always runs.

l0b0 (approved):

blacha (approved):

PR #543, opened 2024-04-17
fix: enable Basemaps preview for Elevation data TDE-1151 BM-932

amfage:
#### Motivation

Enable the Basemaps configuration for `dem_lerc`, as Basemaps now supports previewing Elevation data.

#### Modification

Remove exclusion for `dem_lerc` preset.

#### Checklist

- [ ] Tests updated N/A
- [x] Docs updated
- [x] Issue linked in Title

### DIFF
--- a/workflows/raster/standardising.yaml
+++ b/workflows/raster/standardising.yaml
@@ -377,7 +377,7 @@ spec:
             template: get-location
 
           - name: create-overview
-            when: "'{{workflow.parameters.target_epsg}}' =~ '2193|3857'"
+            when: "'{{workflow.parameters.target_epsg}}' =~ '2193|3857' && '{{workflow.parameters.compression}}' != 'dem_lerc'"
             arguments:
               parameters:
                 - name: location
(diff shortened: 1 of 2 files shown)


l0b0 (commented):
The code looks good, but it would be good to understand why we excluded the DEM LERC files before, and what's changed to make us include them again.

workflows/raster/standardising.yaml
```diff
@@ -377,7 +377,7 @@ spec:
             template: get-location
 
           - name: create-overview
-            when: "'{{workflow.parameters.target_epsg}}' =~ '2193|3857' && '{{workflow.parameters.compression}}' != 'dem_lerc'"
```
blacha, line 380: Overviews do not work with LERC files :sob: , this step still needs to be skipped.

amfage: Oh, so it isn't in production?

amfage: Oh I see, the overviews. Got it.

amfage:
> The code looks good, but it would be good to understand why we excluded the DEM LERC files before, and what's changed to make us include them again.

I have updated the PR description.

paulfouquet (approved):

PR #550, opened 2024-04-23
feat(basemaps): Upgrade vector elt include title in the stac. BM-1017

Wentao-Kuang:
#### Motivation

The new version of vector etl container to which include the fix of road addressing big tiles, and include the title into stac file.

#### Modification

- include title in the etl cmd

#### Checklist

- [ ] Tests updated
- [ ] Docs updated
- [x] Issue linked in Title

### DIFF
--- a/workflows/basemaps/vector-etl.yaml
+++ b/workflows/basemaps/vector-etl.yaml
@@ -17,7 +17,7 @@ spec:
       
       - name: version_basemaps_etl
         description: Version of the Basemaps ETL eks container to use
-        value: 'v1.2'
+        value: 'v1'
 
       - name: target
         description: S3 Bucket to use for storing the output
--- a/workflows/basemaps/vector-etl.yaml
+++ b/workflows/basemaps/vector-etl.yaml
@@ -17,7 +17,7 @@ spec:
       
       - name: version_basemaps_etl
         description: Version of the Basemaps ETL eks container to use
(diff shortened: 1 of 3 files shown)


l0b0 (dismissed):

workflows/basemaps/vector-etl.yaml
```diff
@@ -17,7 +17,7 @@ spec:
       
       - name: version_basemaps_etl
         description: Version of the Basemaps ETL eks container to use
-        value: 'v1'
```
blacha, line 20: we shouldnt need to update this? `v1` should match `v1.x.x` ?

Wentao-Kuang: Yeah, I just prefer to fix to 1.2 first in case any future changes broken the v1 version.

blacha: if there are breaking changes it shouldn't be `v1` it should be made a `v2` ?

workflows/basemaps/vector-etl.yaml
```diff
@@ -17,7 +17,7 @@ spec:
...
+        value: 'v1.2'
```
Wentao-Kuang, line 20:
```suggestion
        value: 'v1'
```

l0b0 (approved):

PR #585, opened 2024-05-15
fix: stac-validate-collections retries should be managed by tpl-at-stac-validate only TDE-1178

paulfouquet:
#### Motivation
In the case of a task in `tpl-at-stac-validate` fails more than twice and the workflow came back failing, the entire `stac-validate-collections` is retried twice which cause to run the inner stac-validate workflow.

#### Modification

Avoid tasks in the `stac-validate-parallel` workflow to retry by default.
Force `aws-list-collections` to retry twice if it fails.
This is a work around as I could not find a way to specify the retryStrategy at a task level using a `templateRef`

#### Checklist

- [ ] Tests updated
- [ ] Docs updated
- [x] Issue linked in Title

### DIFF
--- a/workflows/stac/stac-validate-parallel.yaml
+++ b/workflows/stac/stac-validate-parallel.yaml
@@ -38,6 +38,8 @@ spec:
       image: ''
   templates:
     - name: main
+      retryStrategy:
+        limit: '0' # avoid retrying any of the following task as `tpl-at-stac-validate` already retries its own tasks.
       dag:
         tasks:
           - name: aws-list-collections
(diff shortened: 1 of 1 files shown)


amfage (approved):

l0b0 (approved):
LGTM

PS: `limit` is [defined](https://raw.githubusercontent.com/argoproj/argo-workflows/v3.5.5/api/jsonschema/schema.json) as int or string, so we could use integers here.

paulfouquet:
> LGTM
> 
> PS: `limit` is [defined](https://raw.githubusercontent.com/argoproj/argo-workflows/v3.5.5/api/jsonschema/schema.json) as int or string, so we could use integers here.

Weirdly I've got this warning when I use a integer:
`Incorrect type. Expected "string".yaml-schema: https://raw.githubusercontent.com/argoproj/argo-workflows/v3.5.5/api/jsonschema/schema.json`

l0b0:
> Weirdly I've got this warning when I use a integer: `Incorrect type. Expected "string".yaml-schema: https://raw.githubusercontent.com/argoproj/argo-workflows/v3.5.5/api/jsonschema/schema.json`

```yaml
    "io.k8s.apimachinery.pkg.util.intstr.IntOrString": {
      "type": "string"
    },
```

:facepalm:

PR #604, opened 2024-05-26
fix: Don't override topo-imagery Docker entrypoint for Python TDE-1179

l0b0:
#### Motivation

We use the Docker `ENTRYPOINT` (Argo `command`) to activate the virtualenv before running Python. We've [verified that this change is backwards compatible by running it with topo-imagery v4.7.1](https://argo.linzaccess.com/workflows/argo/test-is-docker-args-w9p88?tab=workflow&nodeId=test-is-docker-args-w9p88-2719834120&nodePanelView=containers&uid=e8af157e-77c5-403c-bbaa-12525efdf7db).

#### Modification

Use `args` rather than `command` for specifying the interpreter and file to run.

#### Checklist

- [ ] Tests updated (N/A)
- [x] Docs updated
- [x] Issue linked in Title

### DIFF
--- a/workflows/test/env.yaml
+++ b/workflows/test/env.yaml
@@ -12,7 +12,6 @@ spec:
     - name: env
       container:
         image: 019359803926.dkr.ecr.ap-southeast-2.amazonaws.com/topo-imagery:latest
-        # Leave `command` if you want to print the environment variables outside the virtualenv, or `args` if you want
+        # Use `command` if you want to print the environment variables outside the virtualenv, or `args` if you want
         # to print the variables inside the virtualenv.
         command: [env]
-        args: [env]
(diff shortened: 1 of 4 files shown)


amfage (approved):

workflows/test/env.yaml
```diff
@@ -12,4 +12,7 @@ spec:
     - name: env
       container:
         image: 019359803926.dkr.ecr.ap-southeast-2.amazonaws.com/topo-imagery:latest
+        # Leave `command` if you want to print the environment variables outside the virtualenv, or `args` if you want
+        # to print the variables inside the virtualenv.
         command: [env]
```
blacha, line 17: does this work with both `args` and `command` defined? should one be commented out?

l0b0: See the comment above the `command` line for an explanation. The idea is that either is valid, but not both.

blacha: it would be good to leave it in a valid state, since this is used for more than just the python side of the fence, `command` would be my preference.

l0b0: Done & documented.

blacha (approved):

amfage (approved):

PR #622, opened 2024-06-07
feat: karpenter allows on-demand instances TDE-1185

paulfouquet:
#### Motivation

Karpenter should allow on-demand instances.

#### Modification

Add `on-demand` in the list of requirements capacity types.

#### Checklist

- [x] Tests updated Tested on nonprod cluster
- [ ] Docs updated
- [x] Issue linked in Title

### DIFF
--- a/infra/charts/karpenter.ts
+++ b/infra/charts/karpenter.ts
@@ -129,7 +129,7 @@ export class KarpenterProvisioner extends Chart {
         // to prevent long running pods (eg kube-dns) being moved.
         taints: [{ key: 'karpenter.sh/capacity-type', value: 'spot', effect: 'NoSchedule' }],
         requirements: [
-          { key: 'karpenter.sh/capacity-type', operator: 'In', values: ['spot'] },
+          { key: 'karpenter.sh/capacity-type', operator: 'In', values: ['spot', 'on-demand'] },
           { key: 'kubernetes.io/arch', operator: 'In', values: ['amd64'] },
           { key: 'karpenter.k8s.aws/instance-family', operator: 'In', values: ['c5', 'c6i', 'c6a'] },
         ],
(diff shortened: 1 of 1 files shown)


infra/charts/karpenter.ts
```diff
@@ -129,7 +129,7 @@ export class KarpenterProvisioner extends Chart {
         // to prevent long running pods (eg kube-dns) being moved.
```
l0b0, line 129: Do we need to update this comment?

infra/charts/karpenter.ts
```diff
@@ -151,7 +151,7 @@ export class KarpenterProvisioner extends Chart {
           { key: 'karpenter.sh/capacity-type', value: 'spot', effect: 'NoSchedule' },
         ],
         requirements: [
-          { key: 'karpenter.sh/capacity-type', operator: 'In', values: ['spot'] },
+          { key: 'karpenter.sh/capacity-type', operator: 'In', values: ['spot', 'on-demand'] },
```
l0b0, line 154: There are 40+ references to the word "spot" in the code base. Do we want to change some of them to "on-demand"?

l0b0 (approved):

infra/charts/karpenter.ts
paulfouquet: I think running "on-demand" instance should be specific (as the cost is different), so if we want to change from "spot" to "on-demand" it should be justified and it is not the goal of this PR.

infra/charts/karpenter.ts
paulfouquet: I don't think we want to change this.

infra/charts/karpenter.ts
```diff
@@ -129,7 +129,7 @@ export class KarpenterProvisioner extends Chart {
         // to prevent long running pods (eg kube-dns) being moved.
         taints: [{ key: 'karpenter.sh/capacity-type', value: 'spot', effect: 'NoSchedule' }],
         requirements: [
-          { key: 'karpenter.sh/capacity-type', operator: 'In', values: ['spot'] },
+          { key: 'karpenter.sh/capacity-type', operator: 'In', values: ['spot', 'on-demand'] },
```
blacha, line 132: this allows the `spot` provisoner type to launch `on-demand` instances, we need a new provisoner specifically for `on-demand`

l0b0 (approved):

blacha (approved):

PR #634, opened 2024-06-17
fix: ensure only argo jobs are run on nodes created by karpenter TDE-1185

blacha:
#### Motivation

We would prefer only short term jobs from argo are run on large nodes that are created by karpenter

#### Modification

taint all provisoners that karpenter has so pods have to opt into being created by karpenter.

#### Checklist

- [ ] Tests updated
- [ ] Docs updated
- [ ] Issue linked in Title


amfage (approved):

l0b0 (approved):
LGTM after skimming https://kubernetes.io/docs/concepts/scheduling-eviction/taint-and-toleration/

PR #697, opened 2024-08-08
feat: exit handler log workflow status in its own property TDE-1231

paulfouquet:
#### Motivation

In order to simplify the Elasticsearch watcher, logging if the workflow status as its own property rather than concatenated in the `msg` makes things easier to handle.

#### Modification

- Add a `workflowStatus` field that takes the workflow status returned by Argo Workflows
- Change the log `msg` to `Workflow:Done`

#### Checklist

- [ ] Tests updated N/A
- [ ] Docs updated N/A
- [x] Issue linked in Title


templates/common/exit.handler.yaml
```diff
@@ -4,7 +4,7 @@ apiVersion: argoproj.io/v1alpha1
 kind: WorkflowTemplate
 metadata:
   # Template to log the workflow status and its parameters.
-  name: tpl-exit-handler
+  name: test-tpl-exit-handler
```
blacha, line 7:
```suggestion
  name: tpl-exit-handler
```

amfage (dismissed):

amfage (approved):

PR #700, opened 2024-08-11
feat: log extra custom key/value pairs onExit TDE-1233

paulfouquet:
#### Motivation

The exit handler logs information that can be used to trigger notifications. A list of key/value pairs can be useful to be able to customize which information can be logged from different workflows.

#### Modification

- Add an input parameter to `tpl-exit-handler` that takes a JSON String object and logs it as `workflowExtras`

#### Checklist

- [ ] Tests updated
- [x] Docs updated
- [x] Issue linked in Title


templates/common/exit.handler.yaml
```diff
@@ -35,6 +38,7 @@ spec:
               level: 20,
               pid: 1,
               msg: 'Workflow:{{inputs.parameters.workflow_status}}',
-              parameters,
```
blacha, line 38:
```
function guessWorkflowGroup() {
if (parameters.source?.startWith('s3://linz-topographic...")) return 'land'
return 'unkown'

}

  ...
  workflowGroup: parameters.source ==== 's3://' ? 'sea' : 'land'
  ...

PR #703, opened 2024-08-13
feat: log workflowGroup from the exit-handler TDE-1230

paulfouquet:
#### Motivation

It is easier to maintain the workflow group (`land` or `sea`) from the exit handler than from elasticsearch watchers.

#### Modification

- log `workflowGroup` based on the source bucket

#### Checklist

- [X] Tests updated
- [X] Docs updated
- [X] Issue linked in Title


paulfouquet:
Converting to draft while adding tests

templates/common/__test__/exit.handler.test.mjs
```diff
@@ -0,0 +1,112 @@
...
+  const action = createTestFunction(ctx);
+
+  const output = {};
+  action();
+
+  return output;
+}
+
+describe('exit handler script template', () => {
+  it('should log workflow status and parameters', () => {
```
blacha, line 35: it would be good to use the built in  test stubber her rather than overwriting the logger

templates/common/exit.handler.yml
```diff
@@ -29,12 +29,27 @@ spec:
           let parameters = {};
           {{= inputs.parameters.workflow_parameters }}.forEach((pair) => (parameters[pair.name] = pair.value));
 
+          function guessWorkflowGroup() {
+            const source = parameters['source'];
+            if (!source) {
```
blacha, line 34: safer to be `typeof source !== 'string'` as `!1` is true then would error the line below with `1.startsWith`

amfage (approved):

PR #776, opened 2024-09-27
docs: Use non-production stack/cluster names for safety TDE-1255

l0b0:
#### Motivation

Make EKS upgrades faster and simpler.

#### Checklist

- [ ] Tests updated (N/A)
- [x] Docs updated
- [x] Issue linked in Title

### DIFF
--- a/docs/infrastructure/kubernetes.version.md
+++ b/docs/infrastructure/kubernetes.version.md
@@ -59,7 +59,9 @@ Below is an example of upgrading from v1.27 to v1.28
 4. Diff the stack to make sure that only versions are updated (change to `Workflows` stack for production)
 
    ```bash
-   npx cdk diff --context=maintainer-arns=$(aws iam list-roles | jq --raw-output '.Roles[] | select(.RoleName | contains("CiTopo")) | select(.RoleName | contains("-CiRole")).Arn'),$(aws iam list-roles | jq --raw-output '.Roles[] | select(.RoleName == "AccountAdminRole").Arn') WorkflowsNP
+   ci_role="$(aws iam list-roles | jq --raw-output '.Roles[] | select(.RoleName | contains("CiTopo")) | select(.RoleName | contains("-CiRole")).Arn')"
+   admin_role="$(aws iam list-roles | jq --raw-output '.Roles[] | select(.RoleName == "AccountAdminRole").Arn')"
+   npx cdk diff --context=maintainer-arns=${ci_role},${admin_role} WorkflowsNP
    ```
    
    The only changes should be Kubernetes version related.
@@ -93,7 +95,7 @@ Below is an example of upgrading from v1.27 to v1.28
 
 <https://docs.aws.amazon.com/eks/latest/userguide/kubernetes-versions.html#version-deprecation>
 > **Are Amazon EKS managed node groups automatically updated along with the cluster control plane version?**
-No. A managed node group creates Amazon EC2 instances in your account. These instances aren't automatically upgraded when you or Amazon EKS update your control plane. For more information, see Updating a managed node group. We recommend maintaining the same Kubernetes version on your control plane and nodes.
+> No. A managed node group creates Amazon EC2 instances in your account. These instances aren't automatically upgraded when you or Amazon EKS update your control plane. For more information, see Updating a managed node group. We recommend maintaining the same Kubernetes version on your control plane and nodes.
 
 This process is necessary to avoid being blocked for a future Kubernetes version upgrade. For example, if Kubernetes get upgraded from `1.27` to `1.28` and the nodes remain in `1.27`, the next time Kubernetes will be upgraded to `1.29`, the upgrade will fail.
 
--- a/docs/infrastructure/kubernetes.version.md
+++ b/docs/infrastructure/kubernetes.version.md
@@ -60,7 +60,7 @@ Below is an example of upgrading from v1.27 to v1.28
 
    ```bash
    ci_role="$(aws iam list-roles | jq --raw-output '.Roles[] | select(.RoleName | contains("CiTopo")) | select(.RoleName | contains("-CiRole")).Arn')"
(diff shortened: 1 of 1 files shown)


docs/infrastructure/kubernetes.version.md
```diff
@@ -56,10 +56,10 @@ Below is an example of upgrading from v1.27 to v1.28
          kubectlLayer: new KubectlV28Layer(this, 'KubeCtlLayer'),
    ```
 
-4. Diff the stack to make sure that only versions are updated
+4. Diff the stack to make sure that only versions are updated (change to `Workflows` stack for production)
 
    ```bash
-   npx cdk diff Workflows -c ci-role-arn=...
+   npx cdk diff --context=maintainer-arns=$(aws iam list-roles | jq --raw-output '.Roles[] | select(.RoleName | contains("CiTopo")) | select(.RoleName | contains("-CiRole")).Arn'),$(aws iam list-roles | jq --raw-output '.Roles[] | select(.RoleName == "AccountAdminRole").Arn') WorkflowsNP
```
blacha, line 62: I'm not sure this command would work with production, would be better to just explain how to get the roles

The CI role should just be fetched from cloudformation

```
aws cloudformation describe-stacks --stack-name=CiTopoProd | jq .Stacks[0].Outputs[0]
```

account admin role is the same in prod/nonprod with `:accountId/role/AccountAdminRole`

workflow maintainer role

```
aws cloudformation describe-stacks --stack-name=TopographicSharedResourcesProd | jq .Stacks[0].Outputs[0]
```

docs/infrastructure/kubernetes.version.md
```diff
@@ -97,23 +97,23 @@ No. A managed node group creates Amazon EC2 instances in your account. These ins
 
 This process is necessary to avoid being blocked for a future Kubernetes version upgrade. For example, if Kubernetes get upgraded from `1.27` to `1.28` and the nodes remain in `1.27`, the next time Kubernetes will be upgraded to `1.29`, the upgrade will fail.
 
-1. Find the nodegroup name for the cluster
+1. Find the nodegroup name for the cluster (change to `Workflows` cluster here and below for production)
```
blacha, line 100: I don't understand the benefit of this change?

docs/infrastructure/kubernetes.version.md
l0b0: > I'm not sure this command would work with production, would be better to just explain how to get the roles

Just tested; both ARN fetchers work in production.

> The CI role should just be fetched from cloudformation

It's much easier to just paste the command.

> account admin role is the same in prod/nonprod with `:accountId/role/AccountAdminRole`

Wouldn't fetching the account ID and then concatenating it would be more complex than the current solution?

> workflow maintainer role
> 
> ```
> aws cloudformation describe-stacks --stack-name=TopographicSharedResourcesProd | jq .Stacks[0].Outputs[0]
> ```

There doesn't seem to be any `TopographicSharedResourcesNonProd` or similar stack in non-prod. Is it only applicable in production, or is it just not needed for this command?

docs/infrastructure/kubernetes.version.md
l0b0: Safety, as mentioned in the commit message. Once the change is ready to go into production we only need to delete two characters from the commands.

docs/infrastructure/kubernetes.version.md
blacha: > it would be more complex than the current solution?

Its super hard to reason about that command as a giant one liner, it needs to be broken down, fetching the outputs via cloudformation is also significantly faster and less error prone than listing roles and globbing for text

> Wouldn't fetching the account ID

fetching the current account id can be done with sts since its all assumed roles

```aws sts get-caller-identity --query Account --output text```

> There doesn't seem to be any TopographicSharedResourcesNonProd or similar stack in non-prod. Is it only applicable in production, or is it just not needed for this command?

It is only applicable to production as its for analysts to connect to the cluster via azuread they do not use a nonprod cluster which is generally deployed per developer and harder to integrate with azuread.

Comment (l0b0) [reply docs/infrastructure/kubernetes.version.md]: I've applied your suggestions to the Bash command, should be a bit faster & easier to grok now.

File: docs/infrastructure/kubernetes.version.md [hunk 3]
```diff
@@ -56,10 +56,13 @@ Below is an example of upgrading from v1.27 to v1.28
          kubectlLayer: new KubectlV28Layer(this, 'KubeCtlLayer'),
    ```
 
-4. Diff the stack to make sure that only versions are updated
+4. Diff the stack to make sure that only versions are updated (change to `Workflows` stack for production)
 
    ```bash
-   npx cdk diff Workflows -c ci-role-arn=...
+   ci_role="$(aws iam list-roles | jq --raw-output '.Roles[] | select(.RoleName | contains("CiTopo")) | select(.RoleName | contains("-CiRole")).Arn')"
```
paulfouquet, line 62: I don't think the AWS topo-li-nonprod "CIRole" should have rights to access to the nonprod cluster since it is not used.

docs/infrastructure/kubernetes.version.md
```diff
@@ -56,10 +56,13 @@ Below is an example of upgrading from v1.27 to v1.28
...
+   admin_role="arn:aws:iam::$(aws sts get-caller-identity --query Account --output text):role/AccountAdminRole"
+   workflow_maintainer_role="$(aws cloudformation describe-stacks --stack-name=TopographicSharedResourcesProd | jq .Stacks[0].Outputs[0])"
```
paulfouquet, line 64: I don't think the maintainers use the nonprod cluster.

paulfouquet (commented):
We currently did not decided as a team if a nonprod cluster should be deployed all the time. This repo does not have a doc to show how to deploy a non prod cluster (the WorkflowNP name is also really arbitrary and is not referenced anywhere).
I feel we should have a decision before modifying this documentation to point to a nonprod cluster.

l0b0:
We decided to not use non-prod for deployment testing.

PR #782, opened 2024-09-29
docs: Use a variable for the node group name TDE-1255

l0b0:
#### Motivation

This way it's easier to run the process without any typos.

#### Modification

Use a variable rather than static values for the node group name.

#### Checklist

- [ ] Tests updated (N/A)
- [x] Docs updated
- [x] Issue linked in Title

### DIFF
--- a/docs/infrastructure/kubernetes.version.md
+++ b/docs/infrastructure/kubernetes.version.md
@@ -100,24 +100,24 @@ This process is necessary to avoid being blocked for a future Kubernetes version
 1. Find the nodegroup name for the cluster
 
    ```bash
-   aws eks list-nodegroups --cluster-name Workflows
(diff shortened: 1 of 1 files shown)


docs/infrastructure/kubernetes.version.md
```diff
@@ -100,24 +100,24 @@ This process is necessary to avoid being blocked for a future Kubernetes version
 1. Find the nodegroup name for the cluster
 
    ```bash
-   aws eks list-nodegroups --cluster-name Workflows
+   node_group_name="$(aws eks list-nodegroups --cluster-name=Workflows | jq --raw-output '.nodegroups[]')"
```
l0b0, line 103: By printing *all* the node group names rather than index 0, we guard against the situation where there's multiple node groups. In that case, they would all be concatenated, so subsequent commands would fail, indicating that a choice must be made.

paulfouquet (approved):

PR #783, opened 2024-09-30
build: Update EKS from 29 to 30 TDE-1255

l0b0:
#### Motivation

Use recent version of AWS EKS.

#### Checklist

- [ ] Tests updated (N/A)
- [x] Docs updated
- [x] Issue linked in Title


amfage (approved):

infra/eks/cluster.ts
```diff
@@ -81,7 +81,7 @@ export class LinzEksCluster extends Stack {
     // TODO: setup up a database CNAME for changing Argo DB without updating Argo config
     // TODO: run a Disaster Recovery test to recover database data
     this.argoDb = new DatabaseInstance(this, ArgoDbInstanceName, {
-      engine: DatabaseInstanceEngine.postgres({ version: PostgresEngineVersion.VER_15_3 }),
+      engine: DatabaseInstanceEngine.postgres({ version: PostgresEngineVersion.VER_15_7 }),
```
blacha, line 84: What process do we have to validate upgrading postgres versions doesnt break anything?

What impact is a minor update? What about major versions?

l0b0: This will not actually upgrade anything - the production version is already at 15.7 due to automatic upgrades.

blacha: do we want to relax this then? could we just do `VER_15` ?

l0b0: I think that might result in deploying a newer version than the one currently in production, if AWS happens to treat `VER_15` as something other than `VER_15_7` at deployment time. Basically, when doing disaster recovery we will need to set this version to the version of the snapshot we want to restore from.

l0b0: > from my understanding, we have auto updating minor versions so it will eventually always be on the same version.

What do you mean? What will be at the same version as what else?

> I have no idea what the impact of V_15 vs V15_7, but with the auto update we the V15_7 is likely wrong very shortly after we deploy it as it will be moved to V15_8?

It's only "wrong" insofar as we need to bump it to the version of current production before applying to production.

> when we do snapshot restores we restore into a new database then load it into the new actual database, so this version means very little aslong as its the same or higher major version for `pg_dump` ?

That's assuming that there's never any issues with dumping and loading into a newer version, which seems risky.

In summary, locking to a major version rather than a minor version would allow us to change the version number in the IaC more rarely, at the cost of potentially risking hard-to-debug production bugs. In conclusion, I prefer to keep it as is.

paulfouquet (approved):

PR #788, opened 2024-10-03
fix: get-location should output the s3 archive workflow directory TDE-1280

paulfouquet:
#### Motivation

Since upgrading our cluster to EKS 1.29, we experience a different behaviour when calling `{{pod.name}}`. This PR changes the way `get-location` build the s3 archive location for the current workflow.

#### Modification

- modify the `get-location` template script behaviour
- refactorisation to allow testing

NOTE: test have been done using `console.log()` because exporting a script containing imports declaration seems very tricky.

#### Checklist

- [x] Tests updated
- [x] Docs updated
- [x] Issue linked in Title


templates/common/__test__/function.helper.ts
```diff
@@ -0,0 +1,40 @@
...
+
+/**
+ * Extract the script from a workflow template file located as its `source`.
+ *
+ * @param path to the workflow template file
+ * @param stopDelimiter optional delimiter to stop the extraction
+ * @returns the script as a string
+ */
+function getFunctionFromScript(path: string, stopDelimiter?: string): string {
+  let func = fs.readFileSync(path, 'utf-8').split('source: |')[1];
```
l0b0, line 11: Do we have a more reliable way to extract data from YAML? Using string parsing is probably going to be quite brittle.

templates/common/__test__/get.location.test.ts
```diff
@@ -0,0 +1,41 @@
...
+
+describe('get-location script template', () => {
+  it('should output workflow artifact location', (t) => {
+    const spy = t.mock.method(console, 'log');
+    runTestFunction(
+      './templates/common/get.location.yml',
+      [
+        {
+          toReplace: '/tmp/',
+          replaceWith: 'memory://tmp/',
```
blacha, line 14: `memory://` is a `fsa` thing and will not write to the correct location here. `/tmp` seems like a decent place to write to,

templates/common/get.location.yml
```diff
@@ -0,0 +1,47 @@
...
+              path: '/tmp/key'
+      script:
+        image: '019359803926.dkr.ecr.ap-southeast-2.amazonaws.com/argo-tasks:{{=sprig.trim(inputs.parameters.version_argo_tasks)}}'
+        command: [node]
+        source: |
+          const argoTemplate = JSON.parse(process.env['ARGO_TEMPLATE']);
+          const podArchiveLoc = argoTemplate.archiveLocation.s3;
+          const workflowArchiveKey = podArchiveLoc.key.substr(0, podArchiveLoc.key.lastIndexOf('/'));
+          const location = `s3://${podArchiveLoc.bucket}/${workflowArchiveKey}/`
+          console.log(`Location: ${location}`);
```
blacha, line 40: this could be logged as JSON ? then these could be mapped/queried later?

```
console.log(JSON.stringify({ location, bucket, key })) 
```

templates/common/get.location.yml
```diff
@@ -0,0 +1,47 @@
...
+              path: '/tmp/bucket'
+          - name: key
+            valueFrom:
+              path: '/tmp/key'
+      script:
+        image: '019359803926.dkr.ecr.ap-southeast-2.amazonaws.com/argo-tasks:{{=sprig.trim(inputs.parameters.version_argo_tasks)}}'
+        command: [node]
+        source: |
+          const argoTemplate = JSON.parse(process.env['ARGO_TEMPLATE']);
+          const podArchiveLoc = argoTemplate.archiveLocation.s3;
```
blacha, line 37: do we handle a trailing `/` on these paths? would `s3://foo/bar/baz/` be treated the same as `s3://foo/bar/baz` and return `s3://foo/bar/`

I like using the URL lib for this `new URL("..", "s3://foo/bar/baz/").href`

templates/common/__test__/function.helper.ts
blacha: could use `yaml` to parse the YAML https://github.com/eemeli/yaml

the `stopDeliminator` seems pretty risky

templates/common/__test__/function.helper.ts
```diff
@@ -0,0 +1,40 @@
...
+ * Create and run a function from a script located in a workflowTemplate file.
+ *
+ * @param workflowPath path to the workflow template file
+ * @param data data to replace in the script
+ * @param stopDelimiter optional delimiter to stop the extraction of the script
+ */
+export function runTestFunction(workflowPath: string, data: FunctionData[], stopDelimiter?: string): void {
+  let func = getFunctionFromScript(workflowPath, stopDelimiter);
+  data.forEach((d) => {
+    func = func.replace(d.toReplace, d.replaceWith);
```
blacha, line 36: you may want `replaceAll` or its only going to replace the first instance.

templates/common/__test__/get.location.test.ts
paulfouquet: I actually forgot to remove this part. 
I had to change the test logic when I went into trouble dealing with imports while building the function from the script. This is actually not doing anything anymore as the script extraction stops before that with the `// Write outputs delimiter.`

templates/common/__test__/function.helper.ts
paulfouquet: > Do we have a more reliable way to extract data from YAML?
could use yaml to parse the YAML https://github.com/eemeli/yaml

I'll have a look at this! thanks

> the stopDeliminator seems pretty risky

It is super risky. I don't know yet how to deal with the imports when the function is created from the script. I will need help on that...

templates/common/get.location.yml
paulfouquet: They are currently accessible as output parameters.
I logged them only for test purpose, but we could look at logging them as JSON.

templates/common/get.location.yml
paulfouquet: > do we handle a trailing / on these paths? 

Places we use the `location` output are not adding the trailing `/` while building the `flat/` directory from it.

blacha: We are going to need to be much more consistent with our URLS, if it has a "/" its a folder if it does not it should be a file, this archive location seems to one more place that a file path is being treated as a folder.

I generally just add a `const targetFolder = f.endsWith("/") ? f : \`${f}/\`` to force a path to end with a `\`

templates/common/__test__/get.location.test.ts
paulfouquet: > /tmp seems like a decent place to write to

I don't really like the idea of actually writting something when executing the tests. Don't you think using the `memory://` could be more appropriate in that case?

templates/common/get.location.yml
paulfouquet: Should we rename the output parameter `targetLocation` to make it more specific that it is a "folder" and not a file?

templates/common/__test__/get.location.test.ts
```diff
@@ -0,0 +1,39 @@
...
+    const location = String(spy.mock.calls[0]?.arguments[0]);
+    assert.equal(location, 'Location: s3://linz-workflows-scratch/2024-10/02-test-get-location-29l4x/');
+
+    const bucket = String(spy.mock.calls[1]?.arguments[0]);
+    assert.equal(bucket, 'Bucket: linz-workflows-scratch');
+
+    const key = String(spy.mock.calls[2]?.arguments[0]);
+    assert.equal(key, 'Key: 2024-10/02-test-get-location-29l4x');
+
+    assert.deepEqual(shimRequired, ['node:fs']);
```
blacha, line 37: this shim is a global constant so this test would fail if it runs twice?

could you pass something to the runTestFunction to trace require calls

it could be a call back for example moving the logic to the test itself.

```typescript
it('should do something', () => {
const required = [];
runTestFunction(...., (req) => {
 required.push(req);
 if (req === 'node:fs') return fs;
 throw new Error('Failed')
})
})
```

paulfouquet: I made the change

templates/common/get.location.yml
paulfouquet: I logged them as JSON

templates/common/__test__/get.location.test.ts
```diff
@@ -0,0 +1,50 @@
...
+                bucket: 'linz-workflows-scratch',
+              },
+            },
+          }),
+        },
+      ],
+      (req) => {
+        shimRequired.push(req);
+        if (req === 'node:fs') return fs;
+        throw new Error('Failed');
```
l0b0, line 29: Should the first line also be conditional on `req === 'node:fs'`? If so, maybe an early exit might be easier to read:
```suggestion
        if (req !== 'node:fs') throw new Error('Failed');
        shimRequired.push(req);
        return fs;
```

templates/common/__test__/get.location.test.ts
```diff
@@ -0,0 +1,50 @@
...
+    );
+    assert.equal(spy.mock.callCount(), 1);
+
+    const logOutputDict = JSON.parse(String(spy.mock.calls[0]?.arguments[0])) as { time: number };
+
+    // override time
+    logOutputDict.time = 1724037007216;
+
+    assert.deepEqual(logOutputDict, {
+      time: 1724037007216,
```
l0b0, line 40: It might be easier (avoiding a magic number) to assert that it's equal to itself rather than overwriting it:

```suggestion
    const logOutputDict = JSON.parse(String(spy.mock.calls[0]?.arguments[0]));

    assert.deepEqual(logOutputDict, {
      time: logOutputDict.time,
```

templates/common/get.location.yml
```diff
@@ -0,0 +1,55 @@
...
+      imagePullPolicy: Always
+      image: ''
+  entrypoint: main
+  templates:
+    - name: main
+      inputs:
+        parameters:
+          - name: version_argo_tasks
+            description: Version of argo-tasks to use
+            default: 'v4'
```
l0b0, line 20: Shouldn't we normally default to `'latest'`?

templates/common/get.location.yml
```diff
@@ -0,0 +1,55 @@
...
+          fs.writeFileSync('/tmp/key', key);
+          console.log(JSON.stringify({
+              time: Date.now(),
+              level: 20,
+              pid: 1,
+              msg: 'Workflow:Location',
+              location,
+              bucket,
+              key,
+            }),)
```
l0b0, line 55: Would it be possible to pull this into a separate file, to make it easier to edit? It doesn't seem to reference any of the workflow properties.

templates/common/get.location.yml
paulfouquet: We usually default to a released tag. Lastest is not safe for production (breaking change car happen for example)

templates/common/get.location.yml
paulfouquet: I could not find a way to have a separate script that can be called from the template. The only ways I could think of are:
- having the script in the container that is used in the template
- generating the workflow file by adding the external script content in the template prior to its deployment into Argo

l0b0 (dismissed):

l0b0 (approved):

PR #792, opened 2024-10-06
feat: create national DEMs workflow TDE-1166

paulfouquet:
#### Motivation

Having a Nationally combined set of DEMs is useful for any data consumer wanted to access to the entire country DEM. This dataset is aimed to be updated on a regular basis to include all new single region/area dataset that is newly published or updated.
image

#### Modification

Add a WorkflowTemplate that create a nationally combined set of DEM based off a configuration file that list the datasets to include in order of priority. This WF template is able to be used to run a cron workflow.

#### Checklist

- [ ] Tests updated
- [x] Docs updated
- [x] Issue linked in Title


paulfouquet:
Moved to https://github.com/linz/topo-workflows/pull/890

PR #796, opened 2024-10-07
test: use yaml parser to extract wf script TDE-1280

paulfouquet:
#### Motivation

Using a yaml parser is safer than splitting on an expected string in the Workflow file.

#### Modification

- integrate https://github.com/eemeli/yaml library
- modify the way to extract a script from a task within a workflow file

#### Checklist

- [x] Tests updated
- [x] Docs updated
- [x] Issue linked in Title


l0b0 (approved):

templates/common/__test__/exit.handler.test.ts
```diff
@@ -2,18 +2,45 @@ import assert from 'node:assert';
...
 function runTestFunction(ctx: { workflowParameters: string; workflowStatus: string }): void {
-  const func = fs.readFileSync('./templates/common/exit.handler.yml', 'utf-8').split('source: |')[1];
-  if (!func) {
-    throw new Error('No script found in the workflow');
-  }
-  const newFunc = func
+  const wfRaw = fs.readFileSync('./templates/common/exit.handler.yml', 'utf-8');
+  const wfTemplate = YAML.parse(wfRaw) as WorkflowTemplate;
+  const script = getScript(wfTemplate, 'main');
+  const newFunc = script
```
blacha, line 43: why does this need to be reassigned?

templates/common/__test__/exit.handler.test.ts
```diff
@@ -2,18 +2,45 @@ import assert from 'node:assert';
...
+ *
+ * @param workflow Workflow template
+ * @param taskName the task to extract the script from
+ * @returns the script of the task
+ */
+function getScript(workflow: WorkflowTemplate, taskName: string): string {
+  if (workflow.spec && workflow.spec.templates) {
+    for (const template of workflow.spec.templates) {
+      if (template.name === taskName) {
+        if (!template.script?.source) {
```
blacha, line 23: it is generally safer to use `obj == null` over `!obj` 

this function could be restructured to reduce indentation too

How about something like:

```typescript

const template = workflow?.spec?.templates?.find(f => f.name === taskName)
if (template  == null) throw new Error(`Task ${taskName} not found in the workflow`);

const source = template.script?.source
if (source == null) throw new Error(`Task ${taskName} has no script`);
return source
```

templates/common/__test__/exit.handler.test.ts
l0b0: It continues on the next line…

blacha: ahh, might be helpful to move the comment up a line then.

l0b0 (approved):

blacha (approved):

PR #841, opened 2024-11-03
fix: create config url should be stored in /flat TDE-1321

paulfouquet:
#### Motivation

The key output by get-location does not end with a trailing `/`. The create-config task should manage the path within the bucket correctly.

#### Modification

Add a `/` between the key and the `flat` in the artifact path for the config-url.

#### Checklist

- [ ] Tests updated
- [ ] Docs updated
- [ ] Issue linked in Title

### DIFF
--- a/workflows/raster/standardising.yaml
+++ b/workflows/raster/standardising.yaml
@@ -394,7 +394,7 @@ spec:
                 - name: collection-id
                   value: '{{tasks.collection-id-setup.outputs.parameters.collection-id}}'
                 - name: target
-                  value: '{{tasks.get-location.outputs.parameters.location}}flat/'
+                  value: '{{=sprig.trimSuffix("/", tasks["get-location"].outputs.parameters.location)}}/flat/'
               artifacts:
                 - name: group_data
                   from: '{{ tasks.group.outputs.artifacts.output }}'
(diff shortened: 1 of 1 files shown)


l0b0 (approved):

amfage (dismissed):

workflows/raster/standardising.yaml
```diff
@@ -650,7 +650,7 @@ spec:
             path: '/tmp/cogify/config-url'
             s3:
               bucket: '{{inputs.parameters.bucket}}'
-              key: '{{inputs.parameters.key}}flat/config-url'
+              key: '{{inputs.parameters.key}}/flat/config-url'
```
blacha, line 653: should this other part of the code change too? 
```
          - 'config'
          - '{{ inputs.parameters.location }}flat/'
```

how many other places are expecting this key to be a folder?

paulfouquet: `{{ inputs.parameters.location }}` is already having a trailing `/`. I think "location" speaks well by itself being a path and should have a trailing `/`.

paulfouquet: I agree with you, we need a something to parse and concatenate paths withing workflow files.

paulfouquet: Did few tests using a couple of `sprig` functions:
`{{=sprig.clean(sprig.join("/", [tasks[get-location].outputs.parameters.location, "flat/"]))}}`

Problem is that the path we're trying to clean is a URL and is not handled properly by the `clean` function which removes the double `/` in `s3://` -> `s3:/`

paulfouquet:
@schmidtnz suggested to `trimSuffix` `/` of the location and key paths before concat them to avoid having to know if these path ends with a `/` or not.

l0b0 (approved):

PR #970, opened 2025-02-07
refactor: templates for standardise-validate and create-collection topo-imagery commands TDE-1346

amfage:
#### Motivation

The [standardise-validate](https://github.com/linz/topo-imagery/blob/master/scripts/standardise_validate.py) and [create-collection](https://github.com/linz/topo-imagery/blob/master/scripts/collection_from_items.py) commands need to be used by multiple workflows, as we are now publishing multiple products such as a National DEM and National Hillshades.

#### Modification

Create separate WorkflowTemplates for `standardise-validate` and `create-collection`.
Update current WorkflowTemplates `national-dem` and `imagery-standardising` that use these templates.

#### Checklist

- [ ] Tests updated
- [x] Docs updated
- [x] Issue linked in Title


paulfouquet (commented):
The `national-dem` WorkflowTemplate is also using this two templates. It would be great to point it to these new templates.

templates/topo-imagery/standardise-validate.yml
```diff
@@ -0,0 +1,121 @@
...
+apiVersion: argoproj.io/v1alpha1
+kind: WorkflowTemplate
+metadata:
+  # Template from https://github.com/linz/topo-imagery
+  # see https://github.com/linz/topo-imagery/blob/master/scripts/standardise_validate.py
+  name: tpl-ti-standardise-validate
+spec:
+  templateDefaults:
+    container:
+      imagePullPolicy: Always
```
schmidtnz, line 12: Not specific to this PR, as this is something we do in most (if not all) our yaml files.
I'm wondering why we specify the `imagePullPolicy` on `templateDefaults` level with empty string for `image` instead of setting the policy where the actual container / image gets set?

amfage: In general, it is in the spec so it applies to all containers within a workflowtemplate.
In this instance, it's just there in case this workflowtemplate is run standalone and not called from another workflowtemplate.

paulfouquet (approved):

templates/topo-imagery/standardise-validate.yml
```diff
@@ -0,0 +1,121 @@
...
+      image: ''
+  entrypoint: main
+  templates:
+    - name: main
+      inputs:
+        parameters:
+          - name: version_topo_imagery
+            description: version of topo-imagery to use
```
schmidtnz, line 20: nit - consistency with other comments (quotes/caps)

templates/topo-imagery/standardise-validate.yml
```diff
@@ -0,0 +1,121 @@
...
+          - '{{inputs.parameters.target}}'
+          - '--preset'
+          - '{{inputs.parameters.compression}}'
+          - '--start-datetime'
+          - '{{=sprig.trim(inputs.parameters.start_datetime)}}'
+          - '--end-datetime'
+          - '{{=sprig.trim(inputs.parameters.end_datetime)}}'
+          - '--collection-id'
+          - '{{inputs.parameters.collection_id}}'
+          - '--create-footprints'
```
schmidtnz, line 99: with all the refactoring happening, would it be worthwhile aligning the name of the parameters of the python script (--create-footprints) with yaml (create_capture_area) ?

templates/topo-imagery/standardise-validate.yml
```diff
@@ -0,0 +1,121 @@
...
+          - '{{inputs.parameters.create_capture_area}}'
```
schmidtnz, line 100: when to `=sprig.trim()` and when not?

amfage: Usually trim input that comes from the user rather than automatically generated or otherwise coded. It isn't consistent though.

templates/topo-imagery/standardise-validate.yml
amfage: Yeah, this was consistent with other templates but it's better to keep it more informative. I will fix up the other templates in another PR.

templates/topo-imagery/create-collection.yaml
```diff
@@ -0,0 +1,149 @@
...
+            archive:
+              none: {}
+
+      container:
+        image: '019359803926.dkr.ecr.ap-southeast-2.amazonaws.com/topo-imagery:{{=sprig.trim(inputs.parameters.version_topo_imagery)}}'
+        args:
+          - python
+          - '/app/scripts/collection_from_items.py'
+          - '--uri'
+          - '{{=sprig.trimSuffix("/", inputs.parameters.location)}}/flat/'
```
schmidtnz, line 109: it might be better to make adding `/flat/` the responsibility of the calling script?

paulfouquet (approved):

PR #986, opened 2025-03-18
build: add yaml linter

blacha:
### Motivation

We are somewhat inconsitent in our yaml styles, some things are quoted where they shouldn't need to be quoted.

This is a first draft of applying a linter, I have tried to use the default rules where they don't conflict with prettier as we already use prettier for the yaml formatting.

### Modifications

Installs `eslint-plugin-yml` yaml linter and adds it to eslint to be run on every pull request

### Verification

build still builds, testing jobs still runs

### DIFF
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -16,6 +16,9 @@ cfg.overrides.push({
     'yml/block-sequence-hyphen-indicator-newline': ['off'],
     // Prettier will sometimes convert these to multiline, so be safe and always use multi
     'yml/block-sequence': ['error', { singleline: 'always' }],
+
+    // Force string values to be quoted
+    'yml/plain-scalar': ['error', 'never', { overrides: { mappingKey: 'always' } }],
   },
   parser: 'yaml-eslint-parser',
 });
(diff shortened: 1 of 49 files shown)


paulfouquet:
We should probably not trust too much the `argo lint --offline templates/ workflows/` and actually test the workflows modified by this PR before merging?

blacha:
> We should probably not trust too much the `argo lint --offline templates/ workflows/` and actually test the workflows modified by this PR before merging?
> 
> > build still builds, testing jobs still runs
> 
> maybe we need to run some of the workflowTemplate too?

I agree, do we have any easy way to run tests off a pull request?

paulfouquet:
> > We should probably not trust too much the `argo lint --offline templates/ workflows/` and actually test the workflows modified by this PR before merging?
> > > build still builds, testing jobs still runs
> > 
> > 
> > maybe we need to run some of the workflowTemplate too?
> 
> I agree, do we have any easy way to run tests off a pull request?

That would be nice to do something like [what have been removed here](https://github.com/linz/topo-workflows/pull/540/files) but deleting the "test" namespace only when the PR is merged so during the time the PR is opened we can test the workflows on that namespace.

paulfouquet (approved):
I don't think it's necessary to apply these changes on the examples in the different readmes, but they may be inconsistent in their format now?

schmidtnz (approved):

paulfouquet (approved):

PR #994, opened 2025-04-14
feat: allow to keep title of existing Collection TDE-1444

paulfouquet:
### Motivation

Some datasets, such as National DEM and National DEM Hillshades, get updated regularly but need their title to remain the same (for example they don't include dates, which are added by the automatic title generation).

### Modifications

- Pass the [`--keep-tile=true` parameter to the create-collection step](https://github.com/linz/topo-imagery/pull/1313)  when called by the `national-dem` and `hillshade` workflows. Default to `--keep-title=false`.

### Verification

Manually ran Workflows.

### DIFF
--- a/templates/topo-imagery/README.md
+++ b/templates/topo-imagery/README.md
@@ -116,6 +116,8 @@ See [collection_from_items.py](https://github.com/linz/topo-imagery/blob/master/
         value: '{{=sprig.trim(workflow.parameters.licensor_list)}}'
       - name: create_capture_dates
         value: 'false'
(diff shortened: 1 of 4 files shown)


schmidtnz:
merge-layers and standardising might need to be updated too?

paulfouquet:
> merge-layers and standardising might need to be updated too?

I assumed they would use the default value which is not keeping the existing title. Happy to force the default if you think it's more clear.

schmidtnz (approved):

PR #995, opened 2025-04-14
feat: add and update workflows to create national DSM TDE-1454

schmidtnz:
### Motivation

As a bulk elevation data user 
I want a nationally combined set of DSMs (Digital Surface Models)
so that I can use one data source to find the latest and greatest elevation data instead of merging together a large number of different surveys.

### Modifications

Updated `national-dem` workflow to support `dsm` and renamed to `national-elevation`.
Updated national DEM cron workflow accordingly.
Added national DSM cron workflow.
Docs updated.

### Verification

Manual run of new cron workflow.


paulfouquet (commented):
You may need to update [this README](https://github.com/linz/topo-workflows/blob/master/workflows/raster/README.md#national-dem) too

paulfouquet (dismissed):

paulfouquet (approved):

PR #1000, opened 2025-04-23
feat: cron workflows for merged hillshades TDE-1428

schmidtnz:
### Motivation

As a Elevation Data Manager
I want the NZ 1m DEM/DSM, NZ 1m DEM/DSM Hillshade and NZ DEM/DSM Hillshade (1m/8m merged where applicable) to be updated automatically when new elevation data is published on the ODR
So that our published data remains in sync

### Modifications

Added cron workflows for the merged (DSM / DEM; Hillshade / Hillshade - Igor) datasets.

### Verification

Manual workflow runs


paulfouquet (commented):
For consistency with other hillshade related cron workflows ([national dem hillshade ](https://github.com/linz/topo-workflows/blob/master/workflows/cron/cron-national-dem-hillshades.yaml) and [national dsm hillshade](https://github.com/linz/topo-workflows/blob/master/workflows/cron/cron-national-dsm-hillshades.yaml)), it would be better to group the different hillshade types (default / igor) into the same cron workflow. What do you think?

schmidtnz:
> For consistency with other hillshade related cron workflows ([national dem hillshade ](https://github.com/linz/topo-workflows/blob/master/workflows/cron/cron-national-dem-hillshades.yaml) and [national dsm hillshade](https://github.com/linz/topo-workflows/blob/master/workflows/cron/cron-national-dsm-hillshades.yaml)), it would be better to group the different hillshade types (default / igor) into the same cron workflow. What do you think?

We can have a more in-depth chat about this. In short, it's a fair bit harder to manage for the merge workflows due to the input they need and would either look messy or be a different workflow for DSM / DEM, at which point this here with individual cron jobs became my preferred option.

workflows/cron/README.md
```diff
@@ -40,11 +40,22 @@ The two cron workflows `cron-national-dem` and `cron-national-dsm` trigger the `
...
 - [New Zealand LiDAR 1m DEM Hillshade](https://github.com/linz/elevation/blob/master/stac/new-zealand/new-zealand/dem-hillshade_1m/2193/collection.json)
 - [New Zealand LiDAR 1m DEM Hillshade - Igor](https://github.com/linz/elevation/blob/master/stac/new-zealand/new-zealand/dem-hillshade-igor_1m/2193/collection.json)
 - [New Zealand LiDAR 1m DSM Hillshade](https://github.com/linz/elevation/blob/master/stac/new-zealand/new-zealand/dsm-hillshade_1m/2193/collection.json)
 - [New Zealand LiDAR 1m DSM Hillshade - Igor](https://github.com/linz/elevation/blob/master/stac/new-zealand/new-zealand/dsm-hillshade-igor_1m/2193/collection.json)
 
 - schedule: **Monday to Friday at 12:30pm**
+
+## National Merged Hillshades
+
+The four cron workflows `cron-national-merged-[dem/dsm]-[hillshade/hillshade-igor]` run on a daily (Mon-Fri) basis to update the below listed merged hillshade datasets after the respective 1m hillshade datasets have been updated:
```
paulfouquet, line 54: Need to update this as there are 2 workflows now

paulfouquet (approved):

PR #1002, opened 2025-04-28
fix: bad variable name `capture_area_dates` TDE-1475

schmidtnz:
### Motivation

National merged hillshade CRON job is failing/

### Modifications

Fixed name of variable `create_capture_dates`


paulfouquet (commented):
I think `create_capture_dates` should be hardcoded (`false` here) as it requires a task to generate this document (as does `mapsheet-coverage`) that is not present here.

paulfouquet (approved):

PR #1007, opened 2025-04-30
feat: add read permissions to new ODR buckets from Argo TDE-1456

amfage:
### Motivation

Give Argo Workflows read access to the two new ODR buckets `nz-coastal` and `nz-topography` in the same way as for `nz-elevation` and `nz-imagery`.

### DIFF
--- a/infra/eks/cluster.ts
+++ b/infra/eks/cluster.ts
@@ -294,9 +294,9 @@ export class LinzEksCluster extends Stack {
      * While those are public buckets, we still need to give permission to Argo
      * as the `--no-sign-request` is not handled in the code.
      */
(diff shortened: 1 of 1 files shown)


infra/eks/cluster.ts
```diff
@@ -296,5 +296,7 @@ export class LinzEksCluster extends Stack {
      */
     Bucket.fromBucketName(this, 'OdrNzImagery', 'nz-imagery').grantRead(argoRunnerSa.role);
```
blacha, line 297: NIT: do we want to sort these alphabetically?

blacha (approved):

infra/eks/cluster.ts
amfage: I've gone with what was the prior convention, which was in order of being provisioned. I don't really have an opinion...

blacha (approved):

PR #1012, opened 2025-05-08
feat: add nz-coastal bucket to workflows TDE-1508

amfage:
### Motivation

- Prepare for publishing to `s3://nz-coastal` on a regular basis as data comes in.
- Now that there is data in `s3://nz-coastal` we want to validate the STAC.

### Modifications

- Add the `nz-coastal` bucket as a destination for ODR publishing.
- Add a workflow step to the STAC validation cron jobs to validate the `nz-coastal` bucket.

### Verification

Manually ran workflow with `nz-coastal` ODR path.


workflows/raster/publish-odr.yaml
```diff
@@ -69,8 +69,10 @@ spec:
         description: 'The ODR bucket name to publish to'
         value: ''
         enum:
-          - 'nz-elevation'
+          - ''
```
paulfouquet, line 72: What is the reason of having two values `''`?

paulfouquet (changes requested):
- https://github.com/linz/topo-workflows/blob/master/workflows/cron/README.md 
- https://github.com/linz/topo-workflows/blob/master/workflows/raster/README.md
would need to be updated :cry:

paulfouquet (approved):

PR #1013, opened 2025-05-12
feat: archive workflow TDE-1484

paulfouquet:
### Motivation
In order to save cloud storage costs, the Data Managers should be able to move source data that have already been processed to a designated archive bucket with a Glacier Deep Archive lifecycle policy. The source data would be deleted and non-current versions expired after a set date (as per source bucket configuration).

#### Features
- Add a `workflowTemplate` `tpl-at-archive-setup` to call the [`linz/argo-tasks/archive-setup`](https://github.com/linz/argo-tasks/pull/1218) command
- Add a `archive` `workflowTemplate` to allow the Data Managers to archive files.
- Give `tpl-at-copy` `workflowTemplate` the ability to compress destination data with `--compress=false/true` option
- Give `tpl-at-copy` `workflowTemplate` the ability to delete source data with `--delete-source=false/true` option

#### Refactor
- Move `copy` workflow into `storage` folder
- Expose all `copy` workflow parameters to its `main` template inputs parameters

### Verification

Argo Workflows ran

### DIFF
--- a/workflows/storage/archive.yaml
+++ b/workflows/storage/archive.yaml
@@ -56,8 +56,14 @@ spec:
           - name: source
             value: '{{workflow.parameters.source}}'
           - name: group
+            value: '{{workflow.parameters.group}}'
+            default: '1000'
           - name: group_size
+            value: '{{workflow.parameters.group_size}}'
+            default: '100Gi'
           - name: version_argo_tasks
+            value: '{{workflow.parameters.version_argo_tasks}}'
+            default: 'v4'
       dag:
         tasks:
           - name: archive-setup
--- a/workflows/storage/archive.yaml
+++ b/workflows/storage/archive.yaml
@@ -37,9 +37,8 @@ spec:
         description: Ticket ID e.g. 'AIP-55'
         value: ''
       - name: source
-        value: 's3://linz-imagery-staging/test/sample/'
-      - name: target
-        value: 's3://linz-imagery-staging/test/sample_target/'
+        description: Path where the files to be archived are located
+        value: ''
       - name: include
         value: '\.tiff?$|\.json$|\.tfw$|/capture-area\.geojson$|/capture-dates\.geojson$'
       - name: exclude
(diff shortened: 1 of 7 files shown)


workflows/storage/copy.yaml
```diff
@@ -115,9 +110,19 @@ spec:
           - name: target
           - name: include
           - name: exclude
+          - name: copy_option
```
schmidtnz, line 113: would it be OK to set a default value of all the input params to the respective workflow param for more flexibility?

schmidtnz (changes requested):

workflows/storage/copy.yaml
paulfouquet: I could not find the documentation about it. This is based on my experience with Argo: when calling the `template` (here `main`) from its `workflowTemplate` (here `copy` - this file) if you don't specify (or hardcode) a `value` in the `template.inputs.parameters.my_param`, it will automatically bind the `workflow.parameters.my_param` (if any) value. (You can see an example with `source` in the `imagery-standardising` `workflowTemplate` - `workflows/raster/standardising.yaml`). Because the official documentation does not seem to be clear about it, we probably should explicitly set the value using the corresponding `workflow.parameters`.

amfage: How does this work with the `default` value also specified (which is used if the parameter isn't provided)? Will there be an error if there is no workflow parameter?

workflows/storage/archive.yaml
```diff
@@ -0,0 +1,120 @@
...
+
+  templates:
+    - name: main
+      retryStrategy:
+        expression: 'false'
+      inputs:
+        parameters:
+          - name: source
+            value: '{{workflow.parameters.source}}'
+          - name: group
```
schmidtnz, line 58: same comment RE workflow parameters as in copy.yaml - should we add defaults?

schmidtnz (changes requested):

schmidtnz (approved):

PR #1015, opened 2025-05-13
feat: compress and delete source files TDE-1482

schmidtnz:
**Motivation**
The requirement to move data to a designated archive bucket with a Glacier Deep Archive lifecycle policy in order to retain data not frequently accessed for the medium/long term at a lower cost. The source data would be deleted and noncurrent versions expired after a set date.

**Modification**

- Give copy workflow the ability to compress destination data with `--compress` option
- Give copy workflow the ability to delete source data with `--delete-source` option

### DIFF
--- a/templates/argo-tasks/copy.yml
+++ b/templates/argo-tasks/copy.yml
@@ -31,6 +31,20 @@ spec:
               - '--force'
               - '--force-no-clobber'
 
+          - name: compress
+            description: Compress files during copy
+            default: '--compress'
+            enum:
+              - '--compress'
+              - ''
+
+          - name: delete_source
+            description: Delete source file after successful copy
+            default: ''
+            enum:
+              - '--delete-source'
+              - ''
+
           - name: aws_role_config_path
             description: s3 URL or comma-separated list of s3 URLs allowing the workflow to write to a target(s)
             default: 's3://linz-bucket-config/config-write.imagery.json,s3://linz-bucket-config/config-write.elevation.json,s3://linz-bucket-config/config-write.topographic.json'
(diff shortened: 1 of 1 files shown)


templates/argo-tasks/copy.yml
```diff
@@ -31,6 +31,20 @@ spec:
               - '--force'
               - '--force-no-clobber'
 
+          - name: compress
+            description: Compress files during copy
+            default: '--compress'
+            enum:
+              - '--compress'
```
paulfouquet, line 38: Using this value does not make it very nice when plugging into the Argo Web UI and is not consistent with the rest of our workflowTemplates

templates/argo-tasks/copy.yml
```diff
@@ -31,6 +31,20 @@ spec:
...
+              - ''
+
+          - name: delete_source
+            description: Delete source file after successful copy
+            default: ''
+            enum:
+              - '--delete-source'
```
paulfouquet, line 45: Using this value does not make it very nice when plugging into the Argo Web UI and is not consistent with the rest of our workflowTemplates

paulfouquet (changes requested):
A couple of comments and similar changes have already been handled in this PR https://github.com/linz/topo-workflows/pull/1013

templates/argo-tasks/copy.yml
blacha: I believe these flags can be set with `--delete-source=true/false` so `--delete-source={{inputs.parameters.delete_source}}`

would that be more consistent?

what downsides do they have with the argo web ui?

paulfouquet: Yes, that's how we generally do it with other flags. I guess in that specific case it may not be exposed to the UI, but not having a consistent pattern makes it harder to use while calling this template from another workflow.

paulfouquet: The downside you would have with the Argo Web UI would be showing a  field with options that would be:
- "flag"
- ""
image
image
which feels a bit "weird" to me

amfage:
Duplicated by https://github.com/linz/topo-workflows/pull/1013

PR #1016, opened 2025-05-19
fix: survey id is not passed to stac-setup TDE-1503

paulfouquet:
### Motivation

Historical imagery dataset have a Historical Survey Number. `stac-setup` needs this information to run when processing a historical dataset. 

### Modifications

- Allow `stac-setup` workflowTemplate to receive the survey ID and pass it to the `linz/argo-tasks` `stac-setup` command
- Pass the Historical Survey Number from the `imagery-standardising` workflow to the `stac-setup` template

### Verification

Argo Workflows test

### DIFF
--- a/workflows/raster/standardising.yaml
+++ b/workflows/raster/standardising.yaml
@@ -3,7 +3,7 @@
 apiVersion: argoproj.io/v1alpha1
 kind: WorkflowTemplate
 metadata:
-  name: imagery-standardising
+  name: test-historical-is
   labels:
     linz.govt.nz/category: raster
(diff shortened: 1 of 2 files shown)


workflows/raster/standardising.yaml
```diff
@@ -3,7 +3,7 @@
 apiVersion: argoproj.io/v1alpha1
 kind: WorkflowTemplate
 metadata:
-  name: imagery-standardising
+  name: test-historical-is
```
blacha, line 6:
```suggestion
  name: imagery-standardising
```

amfage (approved):

PR #1017, opened 2025-05-20
fix: stac-setup is failing when survey_id has no value TDE-1509

paulfouquet:
### Motivation

[linz/argo-tasks](https://github.com/linz/argo-tasks) uses [`cmd-ts`](https://github.com/Schniz/cmd-ts/) which seems to have an issue when an optional parameter is passed with no value at the end of the command.

### Modifications

- conditionally pass the argument
- add documentation for best practices

### Verification

Argo Workflows run

### DIFF
--- a/templates/argo-tasks/stac-setup.yml
+++ b/templates/argo-tasks/stac-setup.yml
@@ -53,10 +53,10 @@ spec:
           - '--end-year={{=sprig.trunc(4, inputs.parameters.end_datetime)}}'
           - '--region={{inputs.parameters.region}}'
           - '--geographic-description={{inputs.parameters.geographic_description}}'
+          - '--survey-id={{inputs.parameters.survey_id}}'
           - '--odr-url={{inputs.parameters.odr_url}}'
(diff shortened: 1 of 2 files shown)


amfage (approved):

templates/argo-tasks/stac-setup.yml
```diff
@@ -53,10 +53,10 @@ spec:
           - '--end-year={{=sprig.trunc(4, inputs.parameters.end_datetime)}}'
           - '--region={{inputs.parameters.region}}'
           - '--geographic-description={{inputs.parameters.geographic_description}}'
+          - '--survey-id={{inputs.parameters.survey_id}}'
```
blacha, line 56: we have a pattern to support optional arguments 

```
- "{{= sprig.empty(inputs.parameters.source_epsg) ? '' : '--source-epsg=' + inputs.parameters.source_epsg }}"
```

blacha (changes requested):

amfage (approved):

blacha (approved):

PR #1022, opened 2025-05-28
ci: deploy workflows in PR namespace on label TDE-1517

paulfouquet:
### Motivation

Deploying test workflows is not straightforward and often requires temporary changes that can be omit to be removed before pushing the changes.

### Modifications

- Add an action that push the changes into a temporary test namespace when a PR label `workflows` is added
- Add an action that delete the namespace on merge or when the label get removed

### Verification

Tests on this PR
Did not test the deletion on merge as the PR needs to be merge
EDIT: deletion of the namespace successful on PR merge


schmidtnz (dismissed):
NIT - the `test-` filenames are a bit misleading... something like `argo-pr-namespace-deploy`/`-remove` might make it clearer that it's not testing but actually adding and deleting namespaces :-)

amfage (approved):

PR #1039, opened 2025-06-25
feat: upgrade cloudflared TDE-1560

paulfouquet:
### Motivation

The current version of `cloudflared` used in production is not working correctly with the version of `karpenter` we want to use (https://github.com/linz/topo-workflows/pull/1036). Also, Cloudflare has release [a Helm chart](https://github.com/cloudflare/helm-charts/) since the last time we deployed the cloudflare-tunnel.

### Modifications

- Create a CDK8S deployment of `cloudflared` using the Helm chart that uses `2024.8.3` of cloudflare-tunnel. (This is not the latest version, the latest is not available from the Helm chart yet - https://github.com/cloudflare/helm-charts/issues/102)

### Verification

Deployed in NonProd


paulfouquet:
The official Helm chart for cloudflared is not well maintained enough.

PR #1041, opened 2025-06-25
feat: add PriorityClass to Cluster TDE-1564

schmidtnz:
### Motivation

Add PriorityClass to the cluster so it can later be applied to Argo server / controller to keep them from getting evicted in favour of workflow runners.

### Modifications

Added PriorityClass definitions.

### Verification

Manual testing on Dev cluster.


infra/constants.ts
```diff
@@ -1,5 +1,5 @@
 /** Cluster name */
-export const ClusterName = 'Workflows';
+export const ClusterName = 'WorkflowsDev';
```
paulfouquet, line 2:
```suggestion
export const ClusterName = 'Workflows';
```

paulfouquet (approved):

PR #1046, opened 2025-06-27
feat: add priority class high-priority to cloudflared TDE-1569

paulfouquet:
### Motivation

Avoid cloudflared pods to be evicted before workflows runner

### Modifications

- add `high-priority` priorityClass to cloudflared pods

### Verification

add the priorityclass to the pod specs


paulfouquet:
> Unsure if `high` or `very high` would be more appropriate here

I think the same as Argo Server would be ideal?

PR #1049, opened 2025-07-04
feat: unarchive workflow TDE-1553

paulfouquet:
### Motivation
Archived data stored in S3 Glacier Deep Archive can not be immediately accessible and requires a restore operation before it can be read. Manual process to do that is not trivial for most of our users.

### Modifications

This change introduces a workflowTemplate that allows users to easily initiate a restore for a set of objects under a specified prefix in an archive bucket. This is the "Part 1" of the unarchive process. A cronWorkflow will be implemented to verify when the files have been actually restored and notify (Slack) users.

#### Workflow

- Add a `unarchive` workflowTemplate that manages simple restoration request (one location, no filters on suffixes).

#### k8s secrets

- Create some secrets to be used by the workflow

#### IAM Roles

- Allow `Workflows-EksWorkflowsArgoRunnerServiceAccountRole` to create S3 Batch Operations jobs
- Allow `Workflows-EksWorkflowsArgoRunnerServiceAccountRole` to pass the S3BatchRestoreRole IAM role

### Verification

Test workflows

### DIFF
--- a/workflows/storage/unarchive.yaml
+++ b/workflows/storage/unarchive.yaml
@@ -93,7 +93,7 @@ spec:
                 - name: archive_directory
                   value: '{{=sprig.trimSuffix("/",sprig.trimPrefix("/",workflow.parameters.archive_directory))}}'
                 - name: archive_path
-                  value: '{{workflow.parameters.archive_bucket}}/{{=sprig.trimSuffix("/",sprig.trimPrefix("/",workflow.parameters.archive_directory))}}'
+                  value: '{{workflow.parameters.archive_bucket}}/{{=sprig.trimSuffix("/",sprig.trimPrefix("/",sprig.trim(workflow.parameters.archive_directory)))}}'
                 - name: retrieval_tier
                   value: '{{= workflow.parameters.urgent == "true" ? "STANDARD" : "BULK"}}'
                 - name: manifest_list


infra/eks/cluster.ts
```diff
@@ -290,6 +290,21 @@ export class LinzEksCluster extends Stack {
...
+        actions: ['s3:CreateJob'],
+        resources: [`arn:aws:s3:${this.region}:${this.account}:job/*`],
+      }),
+    );
+
+    // give permission to the sa to pass the S3 Batch Operations job role
+    argoRunnerSa.role.addToPrincipalPolicy(
+      new PolicyStatement({
+        actions: ['iam:PassRole'],
+        resources: [`arn:aws:iam::${this.account}:role/S3BatchRestoreRole`], // FIXME: hardcoded role name
```
amfage, line 305: Was this FIXME intended to be fixed in this PR?

schmidtnz (approved):

workflows/storage/README.md
```diff
@@ -37,9 +37,84 @@ Access permissions are controlled by the [Bucket Sharing Config](https://github.
...
+1. The user submit an `unarchive` workflow specifying:
+
+   - the archive bucket where the s3 objects (_files_) need to be restored from. Example: `linz-hydrographic-archive`
+   - the path to the S3 objects within the archive bucket. Example: TODO
+   - a title for the unarchive request. TBD with users
+   - a retrieval tier. `BULK` for "normal" request, `STANDARD` for urgent requests.
+
+   The workflow uses the `create-job` operation from `s3control` to request the restoration all objects under a specified prefix within an S3 archive bucket. It leverages the Manifest Generator to dynamically identify the objects to restore. It's a S3 Batch Operations "restore" job that is submitted by the system.
+
+   The user gets notified on Slack that the unarchive request has been initiated.
```
paulfouquet, line 63: RISK: the job submission might have work but the job could fail without the user knowing?

infra/eks/cluster.ts
```diff
@@ -173,7 +178,7 @@ export class LinzEksCluster extends Stack {
 
     new CfnOutput(this, CfnOutputKeys.ClusterEndpoint, { value: this.cluster.clusterEndpoint });
 
-    this.configureEks();
+    this.configureEks(props.s3BatchRestoreRoleArn);
```
schmidtnz, line 181: all other configuration is stored as class properties.
Might be better to keep it consistent for the batch restore ARN and set it that way instead of adding it to the configureEks method signature?

workflows/storage/unarchive.yaml
```diff
@@ -0,0 +1,213 @@
...
+          - 'linz-hydrographic-archive'
+      - name: archive_directory
+        description: 'The directory in the archive bucket where the data to unarchive is stored. Does not include the bucket name. Example: "aerialsurveys/Invercargill2022_Pgrm3016/".'
+        value: ''
+      - name: consumer
+        description: 'The consumer for who the data is being unarchived. The unarchive data will be copied to the consumer shared location.'
+        value: 'linz'
+        enum:
+          - 'linz'
+      - name: urgent
```
schmidtnz, line 51: is this part of the original design? I didn't think we wanted to expose this to the user.

paulfouquet: I personally think the AWS naming BULK / STANDARD tiers is terrible to understand which one is processing the request faster...

paulfouquet: > is this part of the original design? I didn't think we wanted to expose this to the user.

As discussed with @dwsilk it can be useful to have faster requests.

schmidtnz (approved):

schmidtnz (approved):

workflows/storage/unarchive.yaml
```diff
@@ -0,0 +1,213 @@
...
+          - name: create-manifest
+            templateRef:
+              name: tpl-create-manifest
+              template: main
+            arguments:
+              parameters:
+                - name: source
+                  value: 's3://{{workflow.parameters.archive_bucket}}/{{=sprig.trimPrefix("/",workflow.parameters.archive_directory)}}'
+                - name: target
+                  value: 's3://{{= workflow.parameters.archive_bucket == "linz-topographic-archive" ? "linz-topographic-shared" : workflow.parameters.archive_bucket == "linz-hydrographic-archive" ? "linz-hydrographic-shared" : "unknown-bucket" }}/{{workflow.parameters.consumer}}/{{=sprig.trimPrefix("/",workflow.parameters.archive_directory)}}'
```
amfage, line 74: This results in a duplicated consumer in the path e.g.
`s3://linz-topographic-shared/skyvuw/skyvuw/SN12545A/TILES/SN12545A_CF10_10k_0303.prj`

workflows/storage/unarchive.yaml
```diff
@@ -0,0 +1,213 @@
...
+          - name: restore
+            template: restore
+            arguments:
+              parameters:
+                - name: archive_bucket
+                  value: '{{workflow.parameters.archive_bucket}}'
+                - name: archive_directory
+                  value: '{{=sprig.trimSuffix("/",sprig.trimPrefix("/",workflow.parameters.archive_directory))}}'
+                - name: archive_path
+                  value: '{{workflow.parameters.archive_bucket}}/{{=sprig.trimSuffix("/",sprig.trimPrefix("/",workflow.parameters.archive_directory))}}'
```
amfage, line 96:
```suggestion
                  value: '{{workflow.parameters.archive_bucket}}/{{=sprig.trimSuffix("/",sprig.trimPrefix("/",sprig.trim(workflow.parameters.archive_directory)))}}'
```
in case of white space from copy & paste

workflows/storage/unarchive.yaml
paulfouquet: This risk has been identified and we concluded that a user won't unarchive the entire directory of a supplier.

paulfouquet: As discussed we can leave the supplier in the path even if the consumer is the same name

workflows/storage/README.md
amfage: Suggest we monitor this and see if this is a common occurrence and we can address it if it is a problem?

PR #1050, opened 2025-07-08
fix: event-exporter config

paulfouquet:
### Motivation

Since the deployment of https://github.com/linz/topo-workflows/pull/1037, event exporter is not loading its config properly. Seems to be an issue with the helm chart default `layout: {}`

### Modifications

- apply the entire configuration

### Verification

deployed
image


amfage (approved):

paulfouquet:
https://github.com/bitnami/charts/issues/34880

PR #1051, opened 2025-07-08
fix: cloudflared version label TDE-1577

amfage:
### Motivation

Update `cloudflared` version label to `2025.6.1` for consistency. Cosmetic change.

### DIFF
--- a/infra/charts/cloudflared.ts
+++ b/infra/charts/cloudflared.ts
@@ -10,7 +10,8 @@ export class Cloudflared extends Chart {
     id: string,
     props: { tunnelId: string; tunnelSecret: string; accountId: string; tunnelName: string } & ChartProps,
   ) {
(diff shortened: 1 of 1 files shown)


infra/charts/cloudflared.ts
```diff
@@ -10,7 +10,7 @@ export class Cloudflared extends Chart {
     id: string,
     props: { tunnelId: string; tunnelSecret: string; accountId: string; tunnelName: string } & ChartProps,
   ) {
-    super(scope, id, applyDefaultLabels(props, 'cloudflared', '2023.8.2', 'tunnel', 'workflows'));
+    super(scope, id, applyDefaultLabels(props, 'cloudflared', '2025.6.1', 'tunnel', 'workflows'));
```
paulfouquet, line 13: I wonder if we should make it a constant to use in both the label and the image version so this issue is less likely to happen again?

PR #1054, opened 2025-07-31
feat: timeout copy workflow after 15 minutes TDE-1593

schmidtnz:
### Motivation

`copy` workflows sometimes get stuck. 
The `tpl-copy` [template](https://github.com/linz/topo-workflows/blob/master/templates/argo-tasks/README.md#argo-taskscopy---tpl-copy) now times out after 15 minutes (using `activeDeadlineSeconds`) and will retry up to 10 times. This is a measure to avoid having to manually interact with the workflows while we investigate the root cause of this issue.

### Modifications

Added `activeDeadlineSeconds` and `retryStrategy` to `tpl-copy` workflow template. 
```yaml
activeDeadlineSeconds: '900' # Run up to 900 secs / 15 minutes
retryStrategy:
  limit: '10' # Retry up to 10 times
  retryPolicy: 'Always'
```

### Verification

Manual workflow runs with failing and succeeding workflows.

### DIFF
--- a/templates/argo-tasks/README.md
+++ b/templates/argo-tasks/README.md
@@ -91,8 +91,17 @@ steps:
 
 ## argo-tasks/copy - `tpl-copy`
 
-Template for copying a manifest of files between two locations.  
+Template for copying a manifest of files between two locations.
 See https://github.com/linz/argo-tasks#copy
+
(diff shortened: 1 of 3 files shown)


paulfouquet (commented):
I wonder if some documentation about your findings on `activeDeadlineSeconds` may be worth being written in https://github.com/linz/topo-workflows/blob/master/docs/retry.md

amfage (approved):

paulfouquet (approved):

PR #1071, opened 2025-09-18
feat: verify restore and copy files TDE-1592

amfage:
### Motivation

As a Data Manager I want to move data from the `linz-topographic-archive` / `linz-hydrographic-archive` bucket to a designated bucket so that I can unarchive data when it is required again

### Modifications

This PR contains a cron job that triggers a copy-restore workflow. This lists outstanding S3 Glacier Batch Restore request manifests (without the suffix `.done`) and uses the Argo Tasks `verify-restore` command to check the status of the files referenced in the restore manifests. Once all the files are restored they are copied to the relevant sharing bucket.

### Verification

Tested on Argo Workflows nonprod and prod environments.

### DIFF
--- a/templates/argo-tasks/list.yaml
+++ b/templates/argo-tasks/list.yaml
@@ -35,10 +35,6 @@ spec:
             description: Group the files into this size per group (will use the value of `group` or `group_size` that is reached first)
             default: '100Gi'
 
-          - name: aws_role_config_path
-            description: s3 URL or comma-separated list of s3 URLs of roles config files allowing the workflow to read from a location
-            default: ''
-
       outputs:
         parameters:
           - name: files
--- a/templates/argo-tasks/list.yaml
+++ b/templates/argo-tasks/list.yaml
@@ -28,11 +28,11 @@ spec:
             default: ''
 
           - name: group
-            description: Group files by this many files (will use the value of `group` or `group_size` that is reached first)
+            description: Group files by this many files (will use the value of `group` or `group_size` that is reached first). Set to -1 to disable.
             default: '1000'
 
           - name: group_size
-            description: Group the files into this size per group (will use the value of `group` or `group_size` that is reached first)
+            description: Group the files into this size per group (will use the value of `group` or `group_size` that is reached first). Set to -1 to disable.
             default: '100Gi'
 
       outputs:
--- a/templates/argo-tasks/list.yaml
+++ b/templates/argo-tasks/list.yaml
@@ -48,18 +48,16 @@ spec:
           - name: AWS_ROLE_CONFIG_PATH
             value: 's3://linz-bucket-config/config.json'
         args:
-          [
-            'list',
-            '--verbose',
-            '--include',
-            '{{=sprig.trim(inputs.parameters.include)}}',
-            '--exclude',
-            '{{=sprig.trim(inputs.parameters.exclude)}}',
-            '--group',
-            '{{=sprig.trim(inputs.parameters.group)}}',
-            '--group-size',
-            '{{=sprig.trim(inputs.parameters.group_size)}}',
-            '--output',
-            '/tmp/file_list.json',
-            '{{=sprig.trim(inputs.parameters.location)}}',
-          ]
+          - 'list'
+          - '--verbose'
+          - '--include'
+          - '{{=sprig.trim(inputs.parameters.include)}}'
+          - '--exclude'
+          - '{{=sprig.trim(inputs.parameters.exclude)}}'
+          - '--group'
+          - '{{=sprig.trim(inputs.parameters.group)}}'
+          - '--group-size'
+          - '{{=sprig.trim(inputs.parameters.group_size)}}'
+          - '--output'
+          - '/tmp/file_list.json'
+          - '{{=sprig.trim(inputs.parameters.location)}}'
--- a/templates/argo-tasks/list.yaml
+++ b/templates/argo-tasks/list.yaml
@@ -1,0 +1,69 @@
+# yaml-language-server: $schema=https://raw.githubusercontent.com/argoproj/argo-workflows/v3.5.5/api/jsonschema/schema.json
+
+apiVersion: argoproj.io/v1alpha1
+kind: WorkflowTemplate
+metadata:
+  # Template for listing files in a S3 location
+  # See https://github.com/linz/argo-tasks#list
+  name: tpl-at-list
+spec:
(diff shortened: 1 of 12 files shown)


templates/argo-tasks/list.yaml
```diff
@@ -0,0 +1,65 @@
...
+          - name: include
+            description: A regular expression to match object path(s) or name(s) to include in the list
+            default: ''
+
+          - name: exclude
+            description: A regular expression to match object path(s) or name(s) to exclude from the list
+            default: ''
+
+          - name: group
+            description: Group files by this many files (will use the value of `group` or `group_size` that is reached first)
```
schmidtnz, line 31:
```suggestion
            description: Group files by this many files (will use the value of `group` or `group_size` that is reached first). Set to -1 to disable.
```

templates/argo-tasks/list.yaml
```diff
@@ -0,0 +1,65 @@
...
+            default: '1000'
+
+          - name: group_size
+            description: Group the files into this size per group (will use the value of `group` or `group_size` that is reached first)
```
schmidtnz, line 35:
```suggestion
            description: Group the files into this size per group (will use the value of `group` or `group_size` that is reached first). Set to -1 to disable.
```

templates/argo-tasks/list.yaml
```diff
@@ -0,0 +1,65 @@
...
+              path: /tmp/file_list.json
+
+      container:
+        image: '019359803926.dkr.ecr.ap-southeast-2.amazonaws.com/argo-tasks:{{=sprig.trim(inputs.parameters.version_argo_tasks)}}'
+        imagePullPolicy: Always
+        env:
+          - name: AWS_ROLE_CONFIG_PATH
+            value: 's3://linz-bucket-config/config.json'
+        args:
+          [
```
schmidtnz, line 51: NIT - should we aim for consistency between workflow files? 
Others use 
```
args: 
  - arg1
  - arg2
```
instead of 
```
args:
  [
    arg1,
    arg2,
  ]
```

templates/argo-tasks/verify-restore.yaml
```diff
@@ -0,0 +1,46 @@
...
+          - name: version_argo_tasks
+            description: Version of argo-tasks to use
+            default: 'v4'
+          - name: restore_manifest
+            description: S3 path to the restore manifest file
+            default: ''
+      retryStrategy:
+        retryPolicy: 'OnFailure'
+        backoff:
+          duration: '2m' # wait in case this has run before the batch restore request has not fully completed
```
schmidtnz, line 24:
```suggestion
          duration: '2m' # wait in case this has run before the batch restore request has fully completed
```
?

templates/argo-tasks/verify-restore.yaml
```diff
@@ -0,0 +1,46 @@
...
+            '--output',
+            '/tmp/result.json',
+            '{{inputs.parameters.restore_manifest}}',
+          ]
+      outputs:
+        parameters:
+          - name: result
+            description: Result of the restore verification
+            valueFrom:
+              path: /tmp/result.json
```
schmidtnz, line 46: NIT - `.json` might be misleading here, as this file appears to contain only `true` or `false`.
Also, a more descriptive name than `result` might be appropriate (not 100% happy with but thinking along the lines of `all-files-restored-result` or similar)

workflows/storage/copy-restore.yaml
```diff
@@ -0,0 +1,142 @@
...
+              name: tpl-at-list
+              template: main
+            arguments:
+              parameters:
+                - name: version_argo_tasks
+                  value: '{{inputs.parameters.version_argo_tasks}}'
+                - name: location
+                  value: '{{inputs.parameters.reports_location}}'
+                - name: include
+                  value: 'manifest.json$'
```
schmidtnz, line 50: match whole filename?
```suggestion
                  value: '/manifest.json$'
```

templates/common/read.file.yaml
```diff
@@ -0,0 +1,31 @@
...
+
+apiVersion: argoproj.io/v1alpha1
+kind: WorkflowTemplate
+metadata:
+  # Template to output the content of a file in S3
+  name: tpl-read-file
+spec:
+  templateDefaults:
+    container:
+      imagePullPolicy: Always
```
schmidtnz, line 11: This is also specified on the image below (L28) - where it most likely belongs.

templates/common/read.file.yaml
```diff
@@ -0,0 +1,31 @@
...
+            description: Location of the file to read
+            default: ''
+      outputs:
+        parameters:
+          - name: content
+            valueFrom:
+              path: '/tmp/content'
+      script:
+        image: 'public.ecr.aws/aws-cli/aws-cli:2.27.62'
+        imagePullPolicy: Always
```
schmidtnz, line 28: would this be OK to use a cached image?

workflows/storage/copy-restore.yaml
```diff
@@ -0,0 +1,142 @@
...
+            description: List of restore manifest files to flatten
+            default: ''
+      outputs:
+        parameters:
+          - name: flattened_manifests
+            valueFrom:
+              path: /tmp/flattened_manifests.json
+      container:
+        image: peterevans/curl-jq@sha256:f62dbba08fad486a67d323e75dce757efc2ec55b1ddc2e5b78f174c978c3077c
+        imagePullPolicy: Always
```
schmidtnz, line 82: can this be a cached version? Especially since we pinned a version in the previous line

workflows/storage/copy-restore.yaml
```diff
@@ -0,0 +1,142 @@
...
+          - name: manifests_list
+            description: List of restore manifest files to flatten
+            default: ''
+      outputs:
+        parameters:
+          - name: flattened_manifests
+            valueFrom:
+              path: /tmp/flattened_manifests.json
+      container:
+        image: peterevans/curl-jq@sha256:f62dbba08fad486a67d323e75dce757efc2ec55b1ddc2e5b78f174c978c3077c
```
schmidtnz, line 81: NIT -- ok with the version pin but I'd rather not use arbitrary 3rd party images for something simple like `jq`. 
Also seems surprising that this can't be achieved with a simple-ish sprig function. :-(

templates/common/read.file.yaml
amfage: Thanks, missed taking that out.

PR #1073, opened 2025-09-24
feat: use GSD from tileindex validate and remove user input TDE-1653

schmidtnz:
### Motivation

As an imagery maintainer
I want to make sure the GSD specified in metadata is consistent with the resolution of all TIFF images of a survey
### Modifications
`standardising` workflow uses GSD as determined by `tile-index-validate` (`stac-setup` step now depends on `tile-index-validate`). 
`stac-setup` validates GSD against existing data.

Updated all workflows to include new `validate` parameter for `stac-setup`.
Updated readme files.

### Verification

Argo linter and manual standardising workflow runs with imagery and DEM.

### DIFF
--- a/templates/argo-tasks/stac-setup.yml
+++ b/templates/argo-tasks/stac-setup.yml
@@ -24,6 +24,9 @@ spec:
             default: ''
           - name: gsd
             description: 'Dataset GSD in metres'
+          - name: validate
+            description: 'Validate input GSD against the existing ODR_URL (collection.json or first tiff asset)'
+            default: 'true'
           - name: region
             description: 'Region of the dataset'
           - name: geographic_description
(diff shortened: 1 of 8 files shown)


templates/argo-tasks/stac-setup.yml
```diff
@@ -24,6 +24,9 @@ spec:
             default: ''
           - name: gsd
             description: 'Dataset GSD in metres'
+          - name: validate
+            description: 'Validate input GSD against the existing ODR_URL (collection.json or first tiff asset)'
```
paulfouquet, line 28: Does the existing data has to be in AWS ODR to be verified against?
Would it be less confusing to mention the existing data?

paulfouquet (approved):

PR #1078, opened 2025-10-01
feat: notification handler TDE-1586

amfage:
### Motivation

Create a notification handler workflowtemplate to be able to override the `msg` parameter and also to be able to add in custom parameters as argument parameters so they will be collected for the logs. This can be used in a lifecycle hook for steps and tasks as well as workflow onExit.

### Modifications
Specify the `{{workflow.status}}` and `{{workflow.parameters}}` within the notification handler to simplify its use (this could be done for the standard exit handler, or the custom one could replace the standard exit handler at some point). Create a `msg` parameter so the message can be overridden. Any argument parameters specified when calling the template will be included in the logs.

### Verification
Tested in nonprod and on a test standardising workflow in prod. Unit test created.

### DIFF
--- a/templates/common/README.md
+++ b/templates/common/README.md
@@ -145,7 +145,9 @@ The script run by this template is generating a log, including the status of the
 
 ### Template usage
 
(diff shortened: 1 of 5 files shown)


amfage:
change of approach

PR #1188, opened 2026-04-14
feat(topographic-system): Update the map produce workflow to use stac-push. BM-1624

Wentao-Kuang:
### Motivation

We are not ready to use stac-push for the map produce workflow.

### Modifications

- update the `map-produce` workflow to use the stac-push cmd instead of copy, so that we can have the strategy included and upsert the catalogs.
- Delete duplicated `sheet-produce` workflow, which is same as `map-produce` workflow when set `produce_all=false`

### Verification
Push with latest: https://argo.linzaccess.com/workflows/argo/test-topographic-map-produce-wentao-pr-119-fstcz?tab=workflow&nodeId=test-topographic-map-produce-wentao-pr-119-fstcz-2548170503&nodePanelView=containers&uid=e45bf78d-81e1-4395-bd91-0fa63bf6f1af

Push with date only
https://argo.linzaccess.com/workflows/argo/test-topographic-map-produce-wentao-pr-119-fstcz?tab=workflow&uid=e45bf78d-81e1-4395-bd91-0fa63bf6f1af


workflows/topographic-system/map-produce.yaml
```diff
@@ -219,7 +225,7 @@ spec:
           - name: cover-items
             path: /tmp/produce/cover-items.json
       container:
-        image: 725496895483.dkr.ecr.ap-southeast-2.amazonaws.com/github/linz/topographic-system/map:{{ inputs.parameters.version_map_cli }}
```
blacha, line 222: We should be using the github pull through cache if we can.

workflows/topographic-system/map-produce.yaml
```diff
@@ -257,6 +264,35 @@ spec:
...
+            value: s3://linz-bucket-config/config-write.topography.json
+        command: [bash, -c]
+        args:
+          - |
+            node /app/index.cjs stac-push \
+              --source={{ inputs.parameters.source }}catalog.json \
+              --target={{ inputs.parameters.target }} \
+              --category=product \
+              --strategy=date \
+              $( [ "{{ inputs.parameters.latest }}" = "true" ] && echo "--strategy=latest" ) \
```
blacha: It would be best if we could use sprig to format the args so we can avoid the `bash -c`

Im not entirely sure what the sprig logic looks like but hopefully something like
```
{{= inputs.parameters.latest == "true" ?  "--strategy=latest" : "" }}
```

blacha (dismissed):

blacha (approved):

PR #1199, opened 2026-05-05
fix(unarchive): handle spaces correctly in S3 object filters TDE-1904

paulfouquet:
### Motivation

When filtering files, those containing spaces were being silently skipped because `create-manifest` evaluates regexes against URL-encoded S3 keys (where spaces are %20), but our workflow passed literal spaces to the regex.

### Modifications

Updated the include regex using sprig.replace to convert spaces into (%20| ), allowing the regex to match both literal and URL-encoded spaces.

### Verification

Unarchive workflow ran and restore in progress


amfage (approved):

paulfouquet:
Waiting for the file restore workflow to pick up the request and verify everything works fine before merging this.

PR #1247, opened 2026-07-30
feat: target epsg for tileindex-validate TDE-1990

amfage:
### Motivation & Modifications

Pass `target_epsg` to `tileindex-validate` to allow overriding the default 2193 NZTM2000 e.g. for processing Chatham Islands datasets.

### Verification

Workflow run.

### DIFF
--- a/workflows/raster/standardising.yaml
+++ b/workflows/raster/standardising.yaml
@@ -572,7 +572,7 @@ spec:
             depends: 'standardise-validate'
 
           - name: create-config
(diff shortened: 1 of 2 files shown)


workflows/raster/standardising.yaml
schmidtnz, line 577: is `3857` correct in here?

amfage: It was there previously, these tasks can technically support EPSG `3857`, even if there isn't anything upstream that handles it.

schmidtnz (approved):

PR #1257, opened 2026-08-31
feat: output datatype at collection level TDE-2031

ctali:
### Motivation

To output data_type as a parameter similar to gsd

### Modifications

Updated a number of files to output data_type in the same way that is done for gsd

Not to be merged until other tasks in the story are completed.


.gitignore
```diff
@@ -1,3 +1,4 @@
+tsconfig.tsbuildinfo
```
amfage: We don't usually get a `tsconfig.tsbuildinfo` file so I think we don't need it in the .gitignore

templates/argo-tasks/tile-index-validate.yml
```diff
@@ -55,6 +55,11 @@ spec:
             valueFrom:
               path: /tmp/tile-index-validate/gsd
 
+          - name: data_type
+            description: 'Data type of the input tiffs'
+            valueFrom:
+              path: /tmp/tile-index-validate/data_type
```
amfage:
```suggestion
              path: /tmp/tile-index-validate/data-type
```

amfage (approved):

templates/argo-tasks/stac-setup.yml
```diff
@@ -57,6 +60,7 @@ spec:
           - "{{= sprig.empty(inputs.parameters.geographic_description) ? '' : '--geographic-description=' + inputs.parameters.geographic_description }}"
           - "{{= sprig.empty(inputs.parameters.survey_id) ? '' : '--survey-id=' + inputs.parameters.survey_id }}"
           - "{{= sprig.empty(inputs.parameters.odr_url) ? '' : '--odr-url=' + inputs.parameters.odr_url }}"
+          - "{{= sprig.empty(inputs.parameters.data_type) ? '' : '--data-type=' + inputs.parameters.data_type }}"
```
schmidtnz, line 63: `data_type` is not an optional argument in `argo_tasks` / `stac.setup.ts` ([here](https://github.com/linz/argo-tasks/pull/1351/changes#diff-79a87e6ac95e0b0952ec7b3953b8a5278d2b1b1b6b6feb893674fcc82548ff94R66)) but we only pass the argument when it is non-empty.

templates/topo-imagery/standardise-validate.yaml
```diff
@@ -120,6 +124,8 @@ spec:
           - '{{=sprig.trim(inputs.parameters.target_epsg)}}'
           - '--gsd'
           - '{{=sprig.trim(inputs.parameters.gsd)}}'
+          - '--data-type'
```
schmidtnz, line 127: opposite case of [this comment](https://github.com/linz/topo-workflows/pull/1257/changes#r4001549748). 
In Python, the arg parser does not accept an empty string (because it uses [choices](https://github.com/linz/topo-imagery/pull/1643/changes#diff-6ed962f6ac3be95a6abf43297c76b648b93709d121b2c10eb9da3eac1075f80cR80)), so this will fail if run with an empty arg as-is?

